PR #2, opened 2021-10-05
Imported ARN parser and test suite

justinboswell:

### DIFF
--- a/include/aws/sdkutils/resource_name.h
+++ b/include/aws/sdkutils/resource_name.h
@@ -1,0 +1,42 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#ifndef AWS_COMMON_RESOURCE_NAME_H
+#define AWS_COMMON_RESOURCE_NAME_H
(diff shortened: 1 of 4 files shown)


include/aws/sdkutils/resource_name.h
```diff
@@ -0,0 +1,44 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#ifndef AWS_COMMON_RESOURCE_NAME_H
+#define AWS_COMMON_RESOURCE_NAME_H
+#pragma once
```
graebm, line 7: trivial: AWS_SDKUTILS_RESOURCE_NAME_H

also, I know pragma once works everywhere, but we haven't been using it in our C code

include/aws/sdkutils/resource_name.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <aws/sdkutils/exports.h>
```
graebm, line 9: trivial: I thought our convention was to always include <aws/libname/libname.h> from every other header in that lib? ... in which case why does exports.h exist ... and I don't know

PR #13, opened 2022-09-08
Endpoints20 parsing

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Endpoints ruleset parsing


include/aws/sdkutils/endpoints/ruleset.h
```diff
@@ -0,0 +1,123 @@
...
+ */
+AWS_SDKUTILS_API const struct aws_string *aws_endpoints_parameter_get_built_in(
+    const struct aws_endpoints_parameter *parameter);
+
+/*
+ * Default string value. Returns AWS_OP_ERR if parameter is not a string.
+ */
+AWS_SDKUTILS_API int aws_endpoints_parameter_get_default_string(
+    const struct aws_endpoints_parameter *parameter,
+    struct aws_string **out_string);
```
graebm, line 43: document that `out_string` will point to a reference, and users should not free it

trivial: maybe have it be `const aws_string **out_string`?

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+ */
+
+#include <aws/common/array_list.h>
+#include <aws/common/hash_table.h>
+#include <aws/common/json.h>
+#include <aws/common/ref_count.h>
+#include <aws/common/string.h>
+#include <aws/sdkutils/endpoints/private/ruleset_types_impl.h>
+
+#define SKIP_DOC true
```
graebm, line 13: what is SKIP_DOC?

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+    struct aws_string **out_string) {
+    AWS_PRECONDITION(parameter);
+    AWS_PRECONDITION(out_string);
+
+    if (parameter->type == AWS_ENDPOINTS_PARAMETER_STRING) {
+        *out_string = parameter->default_value.string;
+        return AWS_OP_SUCCESS;
+    };
+
+    return AWS_OP_ERR;
```
graebm, line 53: need to aws_raise_error(SOMETHING)
you can't just `return AWS_OP_ERR` unless some other function has failed and raise an error

likewise for all these getters

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+}
+
+int aws_endpoints_parameter_get_default_boolean(const struct aws_endpoints_parameter *parameter, bool *out_bool) {
+    AWS_PRECONDITION(parameter);
+    AWS_PRECONDITION(out_bool);
+
+    if (parameter->type == AWS_ENDPOINTS_PARAMETER_STRING) {
```
graebm, line 60: type == AWS_ENDPOINTS_PARAMETER_~STRING~BOOLEAN

include/aws/sdkutils/endpoints/ruleset.h
```diff
@@ -0,0 +1,123 @@
...
+
+/*
+ * Whether parameter is deprecated.
+ */
+AWS_SDKUTILS_API bool aws_endpoints_parameters_get_is_deprecated(const struct aws_endpoints_parameter *parameter);
+
+/*
+ * Deprecation message. Null if parameter is not deprecated.
+ */
+AWS_SDKUTILS_API const struct aws_string *aws_endpoints_parameter_get_deprecated_message(
```
graebm, line 71: point out that this does NOT raise an error when it returns NULL
since that's common pattern in our APIs and people might expect it

or go ahead and raise an error (not sure exactly what, INVALID_ARGUMENT? something more specific?)

include/aws/sdkutils/endpoints/ruleset.h
```diff
@@ -0,0 +1,123 @@
...
+/*
+ * Decrement ref count
+ */
+AWS_SDKUTILS_API void aws_endpoints_ruleset_release(struct aws_endpoints_ruleset *ruleset);
+
+/*
+ * Get ruleset parameters.
+ */
+AWS_SDKUTILS_API const struct aws_hash_table *aws_endpoints_ruleset_get_parameters(
+    struct aws_endpoints_ruleset *ruleset);
```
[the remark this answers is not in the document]
graebm: also: `const struct aws_endpoints_ruleset *ruleset`

(likewise for all these getters)

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+    bool is_deprecated;
+    struct aws_string *deprecated_message;
+    struct aws_string *deprecated_since;
+};
+
+struct aws_endpoints_ruleset {
+    struct aws_allocator *allocator;
+    struct aws_ref_count ref_count;
+
+    struct aws_array_list rules;
```
graebm, line 53: add comment stating the exact type stored inside`rules`

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+
+    struct aws_string *version;
+    struct aws_string *service_id;
+    struct aws_hash_table *parameters;
```
graebm, line 57: add comment stating key and value type of `parameters`

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+        struct aws_array_list array;
+        struct aws_endpoints_reference *reference;
+        struct aws_endpoints_function *function;
+    } e;
+};
+
+struct aws_endpoints_function {
+    struct aws_allocator *allocator;
+    struct aws_string *fn; /* TODO: precompute hash or enum value to avoid lookup on every eval */
+    struct aws_array_list argv;
```
graebm, line 83: add comments saying what type is in here

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+struct aws_endpoints_fn;
+
+struct aws_endpoints_expr {
+    struct aws_allocator *allocator;
+    enum aws_endpoints_expr_type type;
+    union {
+        struct aws_string *string;
+        int32_t number;
+        bool boolean;
+        struct aws_array_list array;
```
graebm, line 74: add comment for what type is stored in here

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+    struct aws_allocator *allocator;
+    enum aws_endpoints_url_type url_type;
+    union {
+        struct aws_string *template;
+        struct aws_endpoints_reference *reference;
+        struct aws_endpoints_function *function;
+    } url;
+
+    struct aws_string *properties;
+    struct aws_hash_table *headers;
```
graebm, line 96: comment what type is stored in here

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+    union {
+        struct aws_string *template;
+        struct aws_endpoints_reference *reference;
+        struct aws_endpoints_function *function;
+    } error;
+};
+
+struct aws_endpoints_rule_data_tree {
+    struct aws_allocator *allocator;
+    struct aws_array_list rules;
```
graebm, line 111: comment what type is stored in here

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+struct aws_endpoints_condition {
+    struct aws_allocator *allocator;
+    struct aws_endpoints_function *function;
+    struct aws_string *assign;
+};
+
+struct aws_endpoints_rule {
+    struct aws_allocator *allocator;
+
+    struct aws_array_list conditions;
```
graebm, line 123: comment what type is stored in here

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+
+        data_rule->properties = aws_string_new_from_cursor(allocator, &properties_cur);
+    }
+
+    struct aws_json_value *headers_node =
+        aws_json_value_get_from_object(rule_node, aws_byte_cursor_from_c_str("headers"));
+    if (headers_node != NULL) {
+        data_rule->headers = aws_mem_acquire(allocator, sizeof(struct aws_hash_table *));
+        AWS_ZERO_STRUCT(*data_rule->headers);
+        aws_hash_table_init(
```
graebm, line 555: trivial: don't need to zero-out a struct before you call its init() function

you can assume the init() will zero things out for you

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+            goto error_cleanup;
+        }
+
+        data_rule->properties = aws_string_new_from_cursor(allocator, &properties_cur);
+    }
+
+    struct aws_json_value *headers_node =
+        aws_json_value_get_from_object(rule_node, aws_byte_cursor_from_c_str("headers"));
+    if (headers_node != NULL) {
+        data_rule->headers = aws_mem_acquire(allocator, sizeof(struct aws_hash_table *));
```
graebm, line 553: shouldnt' this be `sizeof(struct aws_hash_table)`

alternatively, you could store the aws_hash_tables by-value instead of by-pointer BUT only if you don't mind them moving around in memory .... yeah this is probably more dangerous than it's worth

graebm (commented):
1st blast of feedback
still reading though...

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+
+struct aws_endpoints_expr {
+    struct aws_allocator *allocator;
+    enum aws_endpoints_expr_type type;
+    union {
+        struct aws_string *string;
+        int32_t number;
+        bool boolean;
+        struct aws_array_list array;
+        struct aws_endpoints_reference *reference;
```
graebm, line 75: why not just `struct aws_string *reference;`?

Is there a strong reason to make `aws_endpoints_reference` its own type? Does this introduce some kind of type-safety to the API? Or might we add more members to this struct in the future?

If we keep it as a struct, maybe just store it by-value like:
```c
union {
    ...
    struct aws_endpoints_reference reference;
    ...
}
```
since it's a private-and-not-reference-counted thing and its lifetime is already tied to the `aws_endpoints_expr`? you wouldn't even need to store `aws_endpoints_reference.allocator` member, since its lifetime is 100% tied to the owning `struct aws_endpoints_expr`. You could even just declare the type inline with the declaration of this union like

```c
union {
    struct aws_string *string;
    int32_t number;
    bool boolean;

    struct aws_endpoints_reference {
        struct aws_string *ref;
    } reference;
}
```

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+        struct aws_endpoints_function *function;
+    } e;
+};
+
+struct aws_endpoints_function {
+    struct aws_allocator *allocator;
+    struct aws_string *fn; /* TODO: precompute hash or enum value to avoid lookup on every eval */
```
graebm, line 82: pretty much the same comments as the one about `struct aws_endpoints_reference`

we could just store this by-value within the union, instead of doing another heap allocation, and we probably don't need to store the allocator since we're only destroying this while destroying the `aws_endpoints_expr` that owns it

include/aws/sdkutils/endpoints/ruleset.h
```diff
@@ -0,0 +1,123 @@
...
+#define AWS_SDKUTILS_ENDPOINTS_RULESET_H
+
+#include <aws/common/byte_buf.h>
+#include <aws/sdkutils/sdkutils.h>
+
+struct aws_endpoints_ruleset;
+struct aws_endpoints_parameter;
+struct aws_hash_table;
+
+enum aws_endpoints_parameter_value_type { AWS_ENDPOINTS_PARAMETER_STRING, AWS_ENDPOINTS_PARAMETER_BOOLEAN };
```
graebm, line 16: is there any public doc we can link to that explains what the json file should look like? and what all the fields mean?

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+}
+
+struct aws_endpoints_ruleset *aws_endpoints_ruleset_new_from_string(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *ruleset_cur) {
+    AWS_PRECONDITION(allocator);
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(ruleset_cur));
+
+    struct aws_endpoints_ruleset *ruleset = aws_mem_acquire(allocator, sizeof(struct aws_endpoints_ruleset));
+    AWS_ZERO_STRUCT(*ruleset);
```
graebm, line 857: trivial: use `aws_mem_calloc()` to allocate zeroed-out memory
instead of `aws_mem_acquire(); AWS_ZERO_STRUCT()`

(search through the code for "AWS_ZERO_STRUCT" to find a lot more places you can do this)

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+        struct aws_byte_cursor cur;
+        aws_json_value_get_string(node, &cur);
+        struct aws_endpoints_expr *expr = aws_endpoints_expr_new(allocator, AWS_ENDPOINTS_EXPR_STRING);
+        expr->e.string = aws_string_new_from_cursor(allocator, &cur);
+        return expr;
+    } else if (aws_json_value_is_number(node)) {
+        double number;
+        aws_json_value_get_number(node, &number);
+        struct aws_endpoints_expr *expr = aws_endpoints_expr_new(allocator, AWS_ENDPOINTS_EXPR_NUMBER);
+        expr->e.number = (int32_t)number;
```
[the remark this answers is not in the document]
DmitriyMusatkin: endpoints 2.0 spec does not say what the number is currently. the only place its used is substring function and that defines arg as int and other implementations use int32 as type for all numbers. Im following up on that. Will add comment to clarify that it might change.

include/aws/sdkutils/endpoints/ruleset.h
```diff
@@ -0,0 +1,123 @@
...
+******************************
+* Parameter
+******************************
+*/
+
+/*
+ * Value type of parameter.
+ */
+AWS_SDKUTILS_API enum aws_endpoints_parameter_value_type aws_endpoints_parameter_get_value_type(
+    const struct aws_endpoints_parameter *parameter);
```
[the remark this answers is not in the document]
DmitriyMusatkin: Rulesets are versioned and do not make forwards compat guarantees.
There is a version check in parser (thats currently disabled, because test files are not consistent with versioning). And parsing should fail if it encounters anything thats it does not know about. 
My vote here is to fail parsing and not try to recover, but happy to discuss further.

source/endpoints/ruleset.c
DmitriyMusatkin: removed. left overs from my testing that i forgot to clean up

include/aws/sdkutils/endpoints/ruleset.h
```diff
@@ -0,0 +1,123 @@
...
+
+/*
+ * Decrement ref count
+ */
+AWS_SDKUTILS_API void aws_endpoints_ruleset_release(struct aws_endpoints_ruleset *ruleset);
+
+/*
+ * Get ruleset parameters.
+ */
+AWS_SDKUTILS_API const struct aws_hash_table *aws_endpoints_ruleset_get_parameters(
```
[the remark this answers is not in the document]
DmitriyMusatkin: It will get bound out. We haven't bound out hash table before as far as i can tell. 
Since parameters are just informational structs, i was thinking of just doing a deep copy of the table and creating corresponding params in the binding, which would avoid having to do refcounting on params.

source/endpoints/ruleset.c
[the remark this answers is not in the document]
DmitriyMusatkin: reverted this to be double in parsing and functions can cast it accordingly during eval

source/endpoints/ruleset_types_impl.c
```diff
@@ -0,0 +1,257 @@
...
+
+    aws_mem_release(expr->allocator, expr);
+}
+
+struct aws_endpoints_reference *aws_endpoints_reference_new(struct aws_allocator *allocator, struct aws_string *ref) {
+    AWS_PRECONDITION(ref);
+    struct aws_endpoints_reference *reference = aws_mem_acquire(allocator, sizeof(struct aws_endpoints_reference));
+    AWS_ZERO_STRUCT(*reference);
+    reference->allocator = allocator;
+    reference->ref = ref;
```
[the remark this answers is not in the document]
DmitriyMusatkin: removed reference, since there is no benefit to modeling it for now. replaced with string

include/aws/sdkutils/endpoints/ruleset.h
DmitriyMusatkin: There is no public doc as of right now.

source/endpoints/ruleset.c
DmitriyMusatkin: Whats the general guidance with acquire vs calloc? All of sdkutils is pretty much using acquire

source/endpoints/ruleset.c
```diff
@@ -0,0 +1,878 @@
...
+
+static bool s_on_expr_element(size_t idx, const struct aws_json_value *value_node, void *user_data) {
+    (void)idx;
+    struct array_parser_wrapper *wrapper = user_data;
+    struct aws_endpoints_expr *expr = s_parse_expr(wrapper->allocator, value_node);
+
+    if (expr == NULL) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to parse value.");
+        aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+        return false;
```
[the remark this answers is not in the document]
DmitriyMusatkin: The type was based on what we already used for xml iteration. I've looked at the other places where we are doing foreach kinda loops and its inconsistent how we define iterator function. I do like the suggestion, so pr incoming shortly.

include/aws/sdkutils/endpoints/private/ruleset_types_impl.h
```diff
@@ -0,0 +1,173 @@
...
+
+    struct aws_array_list conditions;
+    struct aws_string *documentation;
+
+    enum aws_endpoints_rule_type type;
+    union {
+        struct aws_endpoints_rule_data_endpoint *endpoint;
+        struct aws_endpoints_rule_data_error *error;
+        struct aws_endpoints_rule_data_tree *tree;
+    } rule_data;
```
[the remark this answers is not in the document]
DmitriyMusatkin: refactored this. most of internal is now by value, where we can get away with it

include/aws/sdkutils/sdkutils.h
```diff
@@ -18,13 +18,17 @@ enum aws_sdkutils_errors {
     AWS_ERROR_SDKUTILS_GENERAL = AWS_ERROR_ENUM_BEGIN_RANGE(AWS_C_SDKUTILS_PACKAGE_ID),
     AWS_ERROR_SDKUTILS_PARSE_FATAL,
     AWS_ERROR_SDKUTILS_PARSE_RECOVERABLE,
+    AWS_ERROR_SDKUTILS_ENDPOINTS_UNSUPPORTED_RULESET,
+    AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED,
+    AWS_ERROR_SDKUTILS_ENDPOINTS_INVALID_ARGUMENT,
```
graebm, line 23: Don't add this, just use `AWS_ERROR_INVALID_ARGUMENT` from aws-c-common.

It's OK to use an error code from a library you depend on

include/aws/sdkutils/endpoints/ruleset.h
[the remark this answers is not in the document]
graebm: Add a comment to this function's docs stating how the bindings should work.

I've [done that before](https://github.com/awslabs/aws-c-io/blob/442f9236db6f4c9b970fe0a08d9cda50bc444a81/include/aws/io/tls_channel_handler.h#L423)

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,935 @@
...
+static int s_parse_endpoints_rule_data_tree(
+    struct aws_allocator *allocator,
+    const struct aws_json_value *rule_node,
+    struct aws_endpoints_rule_data_tree *rule_data) {
+    AWS_PRECONDITION(allocator);
+    AWS_PRECONDITION(rule_node);
+    AWS_PRECONDITION(rule_data);
+
+    struct aws_json_value *rules_node = aws_json_value_get_from_object(rule_node, aws_byte_cursor_from_c_str("rules"));
+    size_t num_rules = aws_json_get_array_size(rules_node);
```
graebm, line 688: need to error-check that `rules_node` is non-null

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,935 @@
...
+        return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+    }
+
+    function->fn = aws_string_new_from_cursor(allocator, &fn_cur);
+
+    struct aws_json_value *argv_node = aws_json_value_get_from_object(node, aws_byte_cursor_from_c_str("argv"));
+    if (argv_node == NULL || !aws_json_value_is_array(argv_node)) {
+        aws_endpoints_function_cleanup(function);
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "No argv or unexpected type.");
+        return AWS_OP_SUCCESS;
```
graebm, line 353: shouldn't this return error?

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,935 @@
...
+        expr->e.boolean = v;
+        return AWS_OP_SUCCESS;
+    } else if (aws_json_value_is_array(node)) {
+        expr->type = AWS_ENDPOINTS_EXPR_ARRAY;
+        size_t num_elements = aws_json_get_array_size(node);
+        aws_array_list_init_dynamic(&expr->e.array, allocator, num_elements, sizeof(struct aws_endpoints_expr));
+        if (s_init_array_from_json(allocator, node, &expr->e.array, s_on_expr_element)) {
+            aws_endpoints_expr_cleanup(expr);
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to parse array value type.");
+            return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
```
graebm, line 300: I'm worried that ALSO we might need to clean up this array if something else goes wrong later in the function

maybe just `goto error` here, and always call `aws_endpoints_expr_cleanup(expr);` from there. That's my general approach: always try to clean up everything

like, what if we parse a list like `"ref": ["i", "am", "not", "a", "ref"]`? We'd succesfully parse the list, but then the call to `s_try_parse_reference()` would fail and we'd leak this memory

DmitriyMusatkin: Parsing array is terminal here and it returns either on success or failure, so failing to parse ref should not leak.
This function already has an on_error, so moved cleanup there.

include/aws/sdkutils/private/endpoints_ruleset_types_impl.h
```diff
@@ -0,0 +1,159 @@
...
+    } rule_data;
+};
+
+struct aws_endpoints_parameter *aws_endpoints_parameter_new(
+    struct aws_allocator *allocator,
+    enum aws_endpoints_parameter_value_type type,
+    const struct aws_byte_cursor *name_cur);
+void aws_endpoints_parameter_destroy(struct aws_endpoints_parameter *parameter);
+
+void aws_endpoints_rule_cleanup(struct aws_endpoints_rule *rule);
```
graebm, line 142: utterly trivial: in our existing codebase it's pretty much always `_clean_up()` not ~_cleanup()~

source/endpoints/ruleset.c
graebm: the general guidance is: use _calloc() if you're allocating a struct, because that avoids having uninitialized memory if you forget to initialize every single variable

use _acquire() if you're allocating something like the backing memory for an aws_byte_buf, where we're not even going to access the memory until something else has overwritten it

graebm (commented):
sorry for being so annoying

mostly just giving "we usually write code exactly like X, even though we failed to formally document X"  advice

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+        struct aws_byte_cursor deprecated_message_cur;
+        struct aws_json_value *deprecated_message_node =
+            aws_json_value_get_from_object(deprecated_node, aws_byte_cursor_from_c_str("message"));
+        if (deprecated_message_node != NULL &&
+            aws_json_value_get_string(deprecated_message_node, &deprecated_message_cur)) {
+            aws_endpoints_parameter_destroy(parameter);
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Unexpected value for deprecated message.");
+            return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+        }
+        parameter->deprecated_message = aws_string_new_from_cursor(wrapper->allocator, &deprecated_message_cur);
```
graebm, line 451: bug:
- if there is no "message" node, then `deprecated_message_cur` is uninitialized
- therefore `parameter->deprecated_message = aws_string_new_from_cursor(wrapper->allocator, &deprecated_message_cur);` could end up as ... anything

probably want to break the if-statement into 2:
```
if deprecated_message_node != NULL:
    if aws_json_value_get_string(deprecated_message_node, &deprecated_message_cur):
        ERROR
    parameter->deprecated_message = aws_string_new_from_cursor(...)

File: source/endpoints_ruleset.c [hunk 32]
```diff
@@ -0,0 +1,939 @@
...
+
+        struct aws_byte_cursor deprecated_since_cur;
+        struct aws_json_value *deprecated_since_node =
+            aws_json_value_get_from_object(deprecated_node, aws_byte_cursor_from_c_str("since"));
+        if (deprecated_since_node != NULL && aws_json_value_get_string(deprecated_since_node, &deprecated_since_cur)) {
+            aws_endpoints_parameter_destroy(parameter);
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Unexpected value for deprecated since.");
+            return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+        }
+        parameter->deprecated_since = aws_string_new_from_cursor(wrapper->allocator, &deprecated_since_cur);
```
graebm, line 461: same bug where `deprecated_since_cur` could be uninitialized data

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+static int s_on_header_element(
+    size_t idx,
+    const struct aws_json_value *value,
+    bool *out_should_continue,
+    void *user_data) {
+    (void)idx;
+    (void)out_should_continue;
+    struct array_parser_wrapper *wrapper = user_data;
+    struct aws_byte_cursor cur;
+    aws_json_value_get_string(value, &cur);
```
graebm, line 514: check and ensure it's actually a string

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+
+static int s_on_headers_key(
+    const struct aws_byte_cursor *key,
+    const struct aws_json_value *value,
+    bool *out_should_continue,
+    void *user_data) {
+    (void)out_should_continue;
+    struct member_parser_wrapper *wrapper = user_data;
+
+    if (aws_json_value_is_array(value)) {
```
graebm, line 529: what if it's NOT an array, should that be a parse error?

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+        size_t num_elements = aws_json_get_array_size(value);
+        struct aws_array_list *headers = aws_mem_calloc(wrapper->allocator, 1, sizeof(struct aws_array_list));
+        aws_array_list_init_dynamic(headers, wrapper->allocator, num_elements, sizeof(struct aws_string *));
+        if (s_init_array_from_json(wrapper->allocator, value, headers, s_on_header_element)) {
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract url.");
+            return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
```
graebm, line 535: failing to clean up `headers` in the case of error.

I'd suggest something like :
```
static int s_on_headers_key(...args...) {
    struct aws_array_list *headers = NULL;
    
    ... rest of code if anything goes wrong then goto error;

    return AWS_OP_SUCCESS;
error:
    if (headers) {
        aws_array_list_deep_cleanup(...)
        aws_mem_release(headers);
    }
    return AWS_OP_ERR;
}

```

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+    struct aws_json_value *properties_node =
+        aws_json_value_get_from_object(rule_node, aws_byte_cursor_from_c_str("properties"));
+    if (properties_node != NULL) {
+
+        struct aws_byte_buf properties_buf;
+        aws_byte_buf_init(&properties_buf, allocator, 0);
+
+        if (aws_byte_buf_append_json_string(properties_node, &properties_buf)) {
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract properties.");
+            goto on_error;
```
graebm, line 591: failing to clean up `properties_buf` if things go wrong

use whatever cleanup strategy you want here, it's a weird tmp variable

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+
+    struct aws_json_value *rules_node = aws_json_value_get_from_object(rule_node, aws_byte_cursor_from_c_str("rules"));
+    if (rules_node == NULL) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Rules node is missing.");
+        return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+    }
+
+    size_t num_rules = aws_json_get_array_size(rules_node);
+    aws_array_list_init_dynamic(&rule_data->rules, allocator, num_rules, sizeof(struct aws_endpoints_rule));
+    if (s_init_array_from_json(allocator, rules_node, &rule_data->rules, s_on_rule_element)) {
```
graebm, line 694: didn't actually check to ensure that `rules_node` is an array

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+
+    enum aws_endpoints_rule_type type;
+    if (aws_byte_cursor_eq_ignore_case(&type_cur, &s_endpoint_type_cur)) {
+        type = AWS_ENDPOINTS_RULE_ENDPOINT;
+    } else if (aws_byte_cursor_eq_ignore_case(&type_cur, &s_error_type_cur)) {
+        type = AWS_ENDPOINTS_RULE_ERROR;
+    } else if (aws_byte_cursor_eq_ignore_case(&type_cur, &s_tree_type_cur)) {
+        type = AWS_ENDPOINTS_RULE_TREE;
+    } else {
+        AWS_FATAL_ASSERT(false);
```
graebm, line 728: ~AWS_FATAL_ASSERT()~ do real error handling here. We're parsing json data

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+    if (s_init_array_from_json(wrapper->allocator, conditions_node, &rule.conditions, s_on_condition_element)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract conditions.");
+        goto error_clean_up;
+    }
+
+    switch (type) {
+        case AWS_ENDPOINTS_RULE_ENDPOINT: {
+            struct aws_json_value *endpoint_node =
+                aws_json_value_get_from_object(value, aws_byte_cursor_from_c_str("endpoint"));
+            if (s_parse_endpoints_rule_data_endpoint(wrapper->allocator, endpoint_node, &rule.rule_data.endpoint)) {
```
graebm, line 755: we need to properly null-check `endpoint_node` here, or in `s_parse_endpoints_rule_data_endpoint()`

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+                AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract endpoint rule data.");
+                goto error_clean_up;
+            }
+            break;
+        }
+        case AWS_ENDPOINTS_RULE_ERROR: {
+            struct aws_json_value *error_node =
+                aws_json_value_get_from_object(value, aws_byte_cursor_from_c_str("error"));
+            if (s_parse_endpoints_rule_data_error(wrapper->allocator, error_node, &rule.rule_data.error)) {
```
graebm, line 764: same, need to null-check `error_node`

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+    struct aws_byte_cursor version_cur;
+    struct aws_json_value *version_node = aws_json_value_get_from_object(root, aws_byte_cursor_from_c_str("version"));
+    if (version_node == NULL || aws_json_value_get_string(version_node, &version_cur)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract version.");
+        aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_UNSUPPORTED_RULESET);
+        goto error_clean_up;
+    }
+
+#ifdef VERSION_CHECK /* TODO: samples are currently inconsistent with versions. skip check for now */
+    if (aws_byte_cursor_eq_c_str(&version_cur, &s_supported_version)) {
```
graebm, line 832: should this be case-sensitive? or not?

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract serviceId.");
+            aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_UNSUPPORTED_RULESET);
+            goto error_clean_up;
+        }
+        ruleset->service_id = aws_string_new_from_cursor(allocator, &service_id_cur);
+    }
+
+    ruleset->parameters = aws_mem_acquire(allocator, sizeof(struct aws_hash_table));
+    aws_hash_table_init(
+        ruleset->parameters,
```
graebm, line 856: trivial:  might be simpler to store the hashtable by-value instead of on the heap? seems like it always exists, so why not?

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+        allocator,
+        20,
+        aws_hash_byte_cursor_ptr,
+        s_byte_cursor_eq,
+        NULL,
+        aws_hash_callback_endpoints_parameter_destroy);
+
+    struct aws_json_value *parameters_node =
+        aws_json_value_get_from_object(root, aws_byte_cursor_from_c_str("parameters"));
+    if (s_init_members_from_json(allocator, parameters_node, ruleset->parameters, s_on_parameter_key)) {
```
graebm, line 866: need to null-check `parameters_node` at some point

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract parameters.");
+        aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+        goto error_clean_up;
+    }
+
+    struct aws_json_value *rules_node = aws_json_value_get_from_object(root, aws_byte_cursor_from_c_str("rules"));
+    if (!aws_json_value_is_array(rules_node)) {
```
graebm, line 873: need to null-check `rules_node` before using it

graebm (commented):
finished `endpoints_ruleset.c`

source/endpoints_ruleset.c
```diff
@@ -0,0 +1,939 @@
...
+    /* required fields */
+    struct aws_byte_cursor type_cur;
+    struct aws_json_value *type_node = aws_json_value_get_from_object(value, aws_byte_cursor_from_c_str("type"));
+    if (type_node == NULL || aws_json_value_get_string(type_node, &type_cur)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "Failed to extract parameter type.");
+        return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED);
+    }
+
+    enum aws_endpoints_value_type type;
+    if (aws_byte_cursor_eq_ignore_case(&type_cur, &s_string_type_cur)) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: models are case sensitive and field names should match exactly as specified, i.e. case sensitive lookups on our side.

source/endpoints_ruleset.c
DmitriyMusatkin: case sensitive is fine for now. version strings will be 2 or 3 numbers separated by dots. so caseness does not matter

graebm (approved):

PR #15, opened 2022-10-07
Endpoint20 eval

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

endpoints eval


source/endpoints_rule_engine.c
```diff
@@ -0,0 +1,2173 @@
+/**
```
TingDaoK, line 1: left from here.

source/partitions.c
```diff
@@ -0,0 +1,263 @@
...
+    struct aws_byte_cursor version_cur;
+    struct aws_json_value *version_node = aws_json_value_get_from_object(root, aws_byte_cursor_from_c_str("version"));
+    if (version_node == NULL || aws_json_value_get_string(version_node, &version_cur)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PARTITIONS_PARSING, "Failed to extract version.");
+        aws_raise_error(AWS_ERROR_SDKUTILS_PARTITIONS_UNSUPPORTED);
+        goto on_error;
+    }
+
+#ifdef ENDPOINTS_VERSION_CHECK /* TODO: samples are currently inconsistent with versions. skip check for now */
+    if (aws_byte_cursor_eq_c_str(&version_cur, &s_supported_version)) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: yeah, this is a bug. fixed

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,292 @@
...
+
+    char country[3] = {0};
+    char location[31] = {0};
+    uint8_t num = 0;
+
+    if (3 == sscanf(copy, "%2[^-]-%30[^-]-%03" SCNu8, country, location, &num)) {
+        for (size_t i = 0; i < sizeof(s_known_countries) / sizeof(s_known_countries[0]); ++i) {
+            if (0 == strncmp(s_known_countries[i], country, 3)) {
+                if (location[0] != 0 && num > 0) {
+                    return aws_byte_cursor_from_c_str("aws");
```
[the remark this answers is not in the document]
DmitriyMusatkin: c string literals are usually allocated off data segment and guaranteed to have program long lifetime

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,292 @@
...
+ * Replaced escaped chars within endpoints templated strings.
+ * Basically replaces {{ with { and }} with }.
+ */
+AWS_SDKUTILS_API int aws_templated_string_strip_replace_escaped(
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor str,
+    struct aws_byte_buf *out_buf) {
+
+    if (aws_byte_buf_init(out_buf, allocator, str.len)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_GENERAL, "Failed to init buffer during str sub");
```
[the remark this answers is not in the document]
DmitriyMusatkin: yeah, naming is off. will refactor a bit and update naming

TingDaoK (commented):
Mostly trivial stuff. I think the work flow is clear.

include/aws/sdkutils/endpoints_rule_engine.h
```diff
@@ -164,7 +166,8 @@ AWS_SDKUTILS_API const struct aws_string *aws_endpoints_ruleset_get_service_id(
  */
 AWS_SDKUTILS_API struct aws_endpoints_rule_engine *aws_endpoints_rule_engine_new(
     struct aws_allocator *allocator,
-    struct aws_endpoints_ruleset *ruleset);
+    struct aws_endpoints_ruleset *ruleset, 
+    struct aws_partitions_config *partitions_config);
```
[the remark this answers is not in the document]
graebm: trivial/debatable: is it likely that more things will factor into making a new `aws_endpoints_rule_engine` in the future? is it worth having an options-struct instead of individual arguments?

if you think it's unlikely, we can leave this as-is, since it seems like options-structs themselves could need a revamp in the near-future

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+/* arbitrary max length of a region. curent longest region name is 16 chars */
+#define AWS_REGION_LEN 50
+
+struct aws_byte_cursor aws_byte_cursor_from_substring(const struct aws_string *src, size_t start, size_t end) {
+    AWS_PRECONDITION(aws_string_is_valid(src));
+    AWS_PRECONDITION(start < end && end <= src->len);
+
+    return aws_byte_cursor_from_array(aws_string_bytes(src) + start, end - start);
+}
+
```
graebm, line 33: trivial: I don't see this function called from anywhere. Delete it if it's not used

include/aws/sdkutils/private/endpoints_eval_util.h
```diff
@@ -0,0 +1,68 @@
...
+/*
+ * Replace escaped chars within endpoints templated strings.
+ * Basically replaces {{ with { and }} with }.
+ * Note: this function does not care about existence of { or } and will
+ * leave them as is.
+ */
+AWS_SDKUTILS_API int aws_templated_string_replace_escaped(
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor str,
+    struct aws_byte_buf *out_buf);
```
graebm, line 26: document that this will initialize `out_buf` if it succeeds, and if it fails then out_buf will remain uninitialized

or you could rename this like an aws_byte_buf_init() function so the lifetimes bit is more clear:

`int aws_byte_buf_init_from_templated_string_replace_escaped(struct aws_byte_buf *dest, struct aws_allocator*, struct aws_byte_cursor src)`

include/aws/sdkutils/private/endpoints_eval_util.h
```diff
@@ -0,0 +1,68 @@
...
+ * Determine whether label is a valid host label.
+ */
+AWS_SDKUTILS_API bool aws_is_valid_host_label(struct aws_byte_cursor label, bool allow_subdomains);
+
+/*
+ * Determines partition from region name.
+ * Note: this basically implements regex-less alternative to regexes specified in
+ * partitions file.
+ */
+AWS_SDKUTILS_API struct aws_byte_cursor aws_map_region_to_partition(struct aws_byte_cursor region);
```
graebm, line 60: Document that it returns an empty cursor if there's no mapping

include/aws/sdkutils/private/endpoints_eval_util.h
```diff
@@ -0,0 +1,68 @@
...
+
+AWS_SDKUTILS_API int aws_uri_normalize_path(struct aws_allocator *allocator, 
+                                            struct aws_byte_cursor path,
+                                            struct aws_byte_buf *out_normalized_path);
+
+AWS_SDKUTILS_API struct aws_string *aws_string_from_json(struct aws_allocator *allocator, struct aws_json_value *value);
```
graebm, line 66: const json value

include/aws/sdkutils/private/endpoints_eval_util.h
```diff
@@ -0,0 +1,68 @@
...
+ * Determines partition from region name.
+ * Note: this basically implements regex-less alternative to regexes specified in
+ * partitions file.
+ */
+AWS_SDKUTILS_API struct aws_byte_cursor aws_map_region_to_partition(struct aws_byte_cursor region);
+
+AWS_SDKUTILS_API int aws_uri_normalize_path(struct aws_allocator *allocator, 
+                                            struct aws_byte_cursor path,
+                                            struct aws_byte_buf *out_normalized_path);
+
```
graebm, line 65: similar to another comment

rename this like an aws_byte_buf_init() function, or clearly document how on success the out-buffer will be initialized, and on failure it will be uninitialized

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -28,84 +22,93 @@ enum aws_endpoints_expr_type {
...
 struct aws_endpoints_expr {
     enum aws_endpoints_expr_type type;
     union {
-        struct aws_string *string;
+        struct aws_byte_cursor string;
         double number;
         bool boolean;
         struct aws_array_list array; /* List of (aws_endpoints_expr) */
-        struct aws_string *reference;
+        struct aws_byte_cursor reference;
```
graebm, line 91: the memory that these cursors are pointing to always outlives the `aws_endpoints_expr`?

and the memory is stable? it's not in an aws_byte_buf that could possibly resize or anything?

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -115,13 +118,13 @@ struct aws_endpoints_rule_data_tree {
...
     struct aws_endpoints_function function;
-    struct aws_string *assign;
+    struct aws_byte_cursor assign;
 };
 
 struct aws_endpoints_rule {
     /* List of (aws_endpoints_condition) */
     struct aws_array_list conditions;
-    struct aws_string *documentation;
+    struct aws_byte_cursor documentation;
```
graebm, line 127: similar question: these cursors are pointing to safe memory that always outlives them?

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+    uint16_t octet[4] = {0};
+    char remainder[2] = {0};
+    if (4 != sscanf(
+                 copy,
+                 IP_CHAR_FMT "." IP_CHAR_FMT "." IP_CHAR_FMT "." IP_CHAR_FMT "%1s",
+                 &octet[0],
+                 &octet[1],
+                 &octet[2],
+                 &octet[3],
+                 remainder)) {
```
graebm, line 51: what is "%1s" remainder?

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+    if (host.len == 0) {
+        return false;
+    }
+
+    if (is_uri_encoded) {
+        if (host.ptr[0] != '[' || host.ptr[host.len - 1] != ']') {
+            return false;
+        }
+        aws_byte_cursor_advance(&host, 1);
+        --host.len;
```
graebm, line 74: if host.len started as 1, and we take 2 away, then host.len will underflow and flip to SIZE_MAX

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+
+    return true;
+}
+
+bool aws_is_ipv6(struct aws_byte_cursor host, bool is_uri_encoded) {
+    if (host.len == 0) {
+        return false;
+    }
+
+    if (is_uri_encoded) {
```
graebm, line 69: why do we only check for "[]" when it's URI-encoded?

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+    for (size_t i = 0; i < 4; ++i) {
+        if (octet[i] > 255) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool aws_is_ipv6(struct aws_byte_cursor host, bool is_uri_encoded) {
```
graebm, line 64: if is_uri_encoded, should we be worrying about ANY character being uri-encoded.
e.g.: "%5B" == '['
???

source/endpoints_eval_util.c
DmitriyMusatkin: will clean up. its victim of refactoring

include/aws/sdkutils/private/endpoints_types_impl.h
DmitriyMusatkin: yes. latest version keeps parsed cjson structure around. all the cursors point to strings inside of that json. lifetime for json is same as ruleset. ruleset keeps a couple byte_bufs around, but cursors never point to them. 
Will add comments on that.

include/aws/sdkutils/private/endpoints_types_impl.h
DmitriyMusatkin: yes

source/endpoints_eval_util.c
DmitriyMusatkin: to catch any garbage after the 4 octets, ex. 1.1.1.1a would match to that %1s and we get 5 matched segments, resulting in false return.

source/endpoints_eval_util.c
DmitriyMusatkin: if check right before will exit if cursor does not start with [ or end with ], so it must have at least 2 chars. point taken though, let me see if i can make it clearer

source/endpoints_eval_util.c
DmitriyMusatkin: standard refers to it as uri encoding, but thats a bit generous. 
all uri encoding means is that ipv6 was wrapped into [] and % transformed into %25. nothing else is encoded and follows regular ipv6 format

source/endpoints_eval_util.c
DmitriyMusatkin: as mentioned in another response, the only char thats allowed to be uri encoded is %. standard does not allow for other chars to be uri encoded and hence we dont support those cases as valid ipv6

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+        }
+    }
+
+    if (2 == sscanf(copy, "us-gov-%30[^-]-%03" SCNu8, location, &num)) {
+        if (location[0] != 0 && num > 0) {
+            return aws_byte_cursor_from_c_str("aws-us-gov");
+        }
+    }
+
+    if (2 == sscanf(copy, "cn-%30[^-]-%03" SCNu8, location, &num)) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: there is an additional check for country codes in the first matcher, so it would match on scanf level, but not on country level and fallthrough to cn match. i kept matching here close to how its defined in partitions regexes, but unifying cn and general checks are probably a cleaner approach

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+     * 2 or more consecutive zero groups can be replaced by double colon (::),
+     *     but only once.
+     * ipv6 literal can be scoped by to zone by appending % followed by zone name
+     * ( does not look like there is length reqs on zone name length. this
+     * implementation enforces that its > 1 )
+     * ipv6 can be embedded in url, in which case it must be wrapped inside []
+     * and % be uri encoded as %25.
+     * Implementation is fairly trivial and just iterates through the string
+     * keeping track of the spec above.
+     */
```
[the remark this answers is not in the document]
DmitriyMusatkin: figuring out regex support is taking a bit and we dont absolutely need regex short term (helper just above workaround all need for regex). So for now we cant use regex for this. Arguably regex based approach will also be super ugly. 
But fair point, will clean it up to make it more readable. Its trying to do all the checks in a single loop without additional memory allocation, so its looking super ugly.

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+    size_t subdomain_count = 0;
+    bool is_valid_host_label = true;
+
+    for (size_t i = 0; i < label.len; ++i) {
+        if (subdomain_count > 63) {
+            is_valid_host_label = false;
+            break;
+        }
+
+        if (label.ptr[i] == '.') {
```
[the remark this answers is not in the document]
DmitriyMusatkin: its iterating one char at the time to confirm that chars are valid. with next split we still need to iterate over chars between splits. 
I can break it into several loops instead, one to check that chars are valid one for splits, etc...

include/aws/sdkutils/endpoints_rule_engine.h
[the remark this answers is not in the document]
DmitriyMusatkin: on const partitions_config i think its worth more discussion. the only reason its not const is because of ref counting. 

Re: wrapping args in options. It's really unlikely that resolution itself will need more arguments in the near future. We did discuss potential needs for caching, but current alignment is that it should happen outside of rule engine, either on client side or crt will introduce a wrapper on top of engine that will provide caching

source/endpoints_eval_util.c
```diff
@@ -0,0 +1,420 @@
...
+     * ipv6 can be embedded in url, in which case it must be wrapped inside []
+     * and % be uri encoded as %25.
+     * Implementation is fairly trivial and just iterates through the string
+     * keeping track of the spec above.
+     */
+
+    uint8_t *zone_delim = memchr(host.ptr, '%', host.len);
+
+    if (zone_delim != NULL) {
+        size_t zone_len = host.len - (size_t)(zone_delim - host.ptr);
```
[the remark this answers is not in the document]
DmitriyMusatkin: aws_byte_cursor_next_split should be enough to address this case.
I misunderstood how it worked from reading docs. i've clarified docs for aws_byte_cursor_next_split in a separate pr.

source/partitions.c
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    if (merge.merge_node != NULL) {
+        partition_info = aws_partition_info_new(wrapper->partitions->allocator, *key);
+        partition_info->info = aws_string_new_from_json_value(wrapper->partitions->allocator, merge.merge_node);
+        aws_json_value_destroy(merge.merge_node);
+    } else {
+        partition_info = aws_partition_info_new(wrapper->partitions->allocator, *key);
+        partition_info->info = wrapper->outputs_str;
+        partition_info->is_copy = true;
```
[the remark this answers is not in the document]
DmitriyMusatkin: wrapper will die, but output_str here is default value for the given partition and has lifetime of partition config.

source/endpoints_rule_engine.c
```diff
@@ -0,0 +1,2190 @@
...
+    struct aws_byte_cursor name,
+    struct aws_byte_cursor value) {
+    AWS_PRECONDITION(allocator);
+
+    struct scope_value *val = s_scope_value_new(allocator, name);
+    val->value.type = AWS_ENDPOINTS_EVAL_VALUE_STRING;
+    val->value.v.string = aws_string_new_from_cursor(allocator, &value);
+    val->value.should_clean_up = false;
+
+    if (aws_hash_table_put(&context->values, &val->name_cur, val, NULL)) {
```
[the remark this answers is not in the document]
TingDaoK: I don't see change made here?

source/endpoints_rule_engine.c
```diff
@@ -0,0 +1,1132 @@
...
+    struct aws_endpoints_request_context *context,
+    struct aws_byte_cursor name,
+    bool value) {
+    AWS_PRECONDITION(allocator);
+
+    struct scope_value *val = s_scope_value_new(allocator, name);
+    val->value.type = AWS_ENDPOINTS_EVAL_VALUE_BOOLEAN;
+    val->value.v.boolean = value;
+
+    if (aws_hash_table_put(&context->values, &val->name_cur, val, NULL)) {
```
TingDaoK, line 736: same here

source/endpoints_rule_engine.c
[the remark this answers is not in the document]
DmitriyMusatkin: Do we want to error out if user overrides the value they provided? im leaning towards no on that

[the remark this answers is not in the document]
TingDaoK: The comments of this function saying `* Adding several value with the same name will result in the error.`. [here](https://github.com/awslabs/aws-c-sdkutils/pull/15/files#diff-80818b2b9a3905a998d50930d22acdc8828886eb64db13a217aa93615c717944R211-R213). 

I don't see there is a requirement from spec. Maybe it's okay to allow duplicate and from rule set [here](https://github.com/awslabs/aws-c-sdkutils/pull/15/files#diff-db43dc5625d7e1b5019b57dc7e315ecefb5fd6fdabb5c7600bf8e693005835c7R481), we also allow duplicate. So, maybe update the doc for the function.

include/aws/sdkutils/private/endpoints_util.h
```diff
@@ -0,0 +1,111 @@
...
+ * Resolve templated string and write it out to buf.
+ * Will parse templated values (i.e. values enclosed in {}) and replace them with
+ * the value returned from resolve_callback.
+ * Note: callback must be able to support syntax for pathing through value (path
+ * provided after #).
+ * Will replace escaped template delimiters ({{ and }}) with single chars.
+ * Supports replacing templated values inside json strings (controlled by
+ * is_json), by ignoring json { and } chars.
+ */
+AWS_SDKUTILS_API int aws_byte_buf_init_from_resolved_templated_string(
```
TingDaoK, line 85: Do we really need the `resolve_callback` to solve the templated values? Can we just pass the scope to the function instead? 

The main reason I suggest is that this approach involves some extra string copy that I think we can avoid. Like: the callback create a string and pass back to the caller to clean up, after copy the value into a buffer seems 

And I don't see any reason using the callback other than taking the scope to fetch the referred value.

source/endpoints_ruleset.c
```diff
@@ -467,32 +463,22 @@ static int s_on_parameter_key(
...
+            goto on_error;
         }
-    } else {
-        parameter->is_deprecated = false;
-        parameter->deprecated_message = NULL;
-        parameter->deprecated_since = NULL;
     }
 
-    aws_hash_table_put(wrapper->table, &parameter->name_cur, parameter, NULL);
+    aws_hash_table_put(wrapper->table, &parameter->name, parameter, NULL);
```
TingDaoK, line 481: trivial: this can fail, but it seems like only fail with OOM, and we don't handle again... Just bring it up, maybe ASSERT on it to be succeed?

tests/CMakeLists.txt
```diff
@@ -67,6 +67,28 @@ add_test_case(resource_name_tostring_failure_test)
...
+add_test_case(test_endpoints_default_values)
+add_test_case(test_endpoints_eventbridge)
+add_test_case(test_endpoints_fns)
+# no associated test cases
+#add_test_case(test_endpoints_get_attr_type_inference) 
+add_test_case(test_endpoints_headers)
+add_test_case(test_endpoints_is_virtual_hostable_s3_bucket)
+#no associated test cases
+#add_test_case(test_endpoints_region_override)
+#add_test_case(test_endpoints_minimal_ruleset)
```
TingDaoK, line 80: why comment them out? What do you mean by `no associated test cases`?

source/endpoints_rule_engine.c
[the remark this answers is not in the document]
DmitriyMusatkin: you are right. doc is outdated. fixed now.

Spec does not really define how the interface between middleware and resolver should look like. It is possible through to have several values with the same name and spec defines precedence on those. Allowing overwrites allows caller to just iterate through sources in reverse precedence order and adding values as it goes along and higher precedence value will be the last one added.

tests/CMakeLists.txt
DmitriyMusatkin: some of the sample rulesets do not have associated test files. updated the test helper to skip resolve part in that case and just validate that parsing suceeds.

source/endpoints_util.c
```diff
@@ -0,0 +1,548 @@
...
+    struct aws_byte_cursor string,
+    aws_endpoints_template_resolve_fn resolve_callback,
+    void *user_data,
+    bool is_json) {
+    AWS_PRECONDITION(allocator);
+
+    struct aws_string *resolved_template = NULL;
+
+    if (aws_byte_buf_init(out_buf, allocator, string.len)) {
+        AWS_ZERO_STRUCT(out_buf);
```
TingDaoK, line 414: super trivial: don't need this, the init will actually do it for you on failure.

source/endpoints_util.c
```diff
@@ -0,0 +1,548 @@
...
+
+    if (s_buf_append_and_update_quote_count(out_buf, split, &quote_count)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_GENERAL, "Failed to append to resolved template buffer.");
+        goto on_error;
+    }
+
+    return AWS_OP_SUCCESS;
+
+on_error:
+    aws_string_destroy(resolved_template);
```
TingDaoK, line 482: should we also clean up the `*out_buf` on error?

source/endpoints_util.c
```diff
@@ -0,0 +1,548 @@
...
+
+on_success:
+    aws_array_list_clean_up(&path_segments);
+    return AWS_OP_SUCCESS;
+
+on_error:
+    aws_array_list_clean_up(&path_segments);
+    *out_value = NULL;
+    return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_FAILED);
+    ;
```
TingDaoK, line 547: trivial: extra ";"

source/partitions.c
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    struct partition_parse_wrapper wrapper = {
+        .outputs_node = outputs_node, .outputs_str = partition_info->info, .partitions = partitions};
+
+    struct aws_json_value *regions_node =
+        aws_json_value_get_from_object(partition_node, aws_byte_cursor_from_c_str("regions"));
+    if (regions_node != NULL && aws_json_const_iterate_object(regions_node, s_on_region_element, &wrapper)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PARTITIONS_PARSING, "Failed to parse regions.");
+        goto on_error;
```
TingDaoK, line 161: this is a bit concern, on this failure, the `partitions->region_to_partition_info` has added the new `partition_info`. Can we move this before the table put? Or, we actually don't care?

source/partitions.c
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    partitions->json_root = root;
+
+    struct aws_byte_cursor version_cur;
+    struct aws_json_value *version_node = aws_json_value_get_from_object(root, aws_byte_cursor_from_c_str("version"));
+    if (version_node == NULL || aws_json_value_get_string(version_node, &version_cur)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PARTITIONS_PARSING, "Failed to extract version.");
+        aws_raise_error(AWS_ERROR_SDKUTILS_PARTITIONS_UNSUPPORTED);
+        goto on_error;
```
TingDaoK, line 189: super trivial: why not just return....

include/aws/sdkutils/private/endpoints_util.h
DmitriyMusatkin: several reasons for why its being passed in as a function:
- all utils avoid knowing anything about resolution types to simplify testing and keep boundary from the engine
- resolving values within the function itself seemed like the function was trying to do too much. right now its just a replace function with a bit of context around special char escaping
- getAttr is relying on the same template resolution logic, but outside of string context.

All of those are arguable. For now i replaced the string copy in resolve with passing owning_cursor, which should get rid of some copies. We still have copies during pathing, but those are hard to avoid, since objects are jsons and there is no way to get at the original string when parsing json.

source/endpoints_standard_lib.c
```diff
@@ -0,0 +1,627 @@
...
+static struct aws_byte_cursor s_scheme_https = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("https");
+
+static int s_resolve_fn_is_set(
+    struct aws_allocator *allocator,
+    struct aws_array_list *argv,
+    struct aws_endpoints_resolution_scope *scope,
+    struct aws_endpoints_value *out_value) {
+
+    struct aws_endpoints_value argv_value;
+    if (aws_endpoints_argv_expect(allocator, scope, argv, 0, AWS_ENDPOINTS_VALUE_ANY, &argv_value)) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: fair point. adding it

source/endpoints_standard_lib.c
```diff
@@ -0,0 +1,627 @@
...
+    struct aws_uri uri;
+    struct aws_json_value *root = NULL;
+    struct aws_endpoints_value argv_url;
+    if (aws_endpoints_argv_expect(allocator, scope, argv, 0, AWS_ENDPOINTS_VALUE_STRING, &argv_url)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE, "Failed to resolve args for parse url.");
+        goto on_error;
+    }
+
+    if (aws_uri_init_parse(&uri, allocator, &argv_url.v.string.cur)) {
+        out_value->type = AWS_ENDPOINTS_VALUE_NONE;
```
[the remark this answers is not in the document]
DmitriyMusatkin: good point. added error reset.

source/endpoints_util.c
```diff
@@ -0,0 +1,548 @@
...
+    return NULL;
+}
+
+bool aws_endpoints_byte_cursor_eq(const void *a, const void *b) {
+    const struct aws_byte_cursor *a_cur = a;
+    const struct aws_byte_cursor *b_cur = b;
+    return aws_byte_cursor_eq(a_cur, b_cur);
+}
+
+void aws_array_list_deep_clean_up(struct aws_array_list *array, aws_array_callback_clean_up_fn on_clean_up_element) {
```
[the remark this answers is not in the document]
DmitriyMusatkin: there is already a todo in the header for it

source/partitions.c
DmitriyMusatkin: dont think its a concern. partitions file has a partition specific meta data section and then each region within partition can override parts of that metadata, which will create a new region specific entry. If region parsing fails, then we partition wide metadata will be cleaned up in the top level clean up

source/endpoints_standard_lib.c
```diff
@@ -225,36 +222,37 @@ static int s_resolve_fn_uri_encode(
     struct aws_endpoints_resolution_scope *scope,
     struct aws_endpoints_value *out_value) {
 
+    int result = AWS_OP_SUCCESS;
     struct aws_byte_buf buf;
```
TingDaoK, line 230: `buf` needs to be zeroed out.

source/endpoints_standard_lib.c
```diff
@@ -0,0 +1,639 @@
...
+
+    int result = AWS_OP_SUCCESS;
+    struct aws_endpoints_value argv_value_1 = {0};
+    struct aws_endpoints_value argv_value_2 = {0};
+    if (aws_array_list_length(argv) != 2 ||
+        aws_endpoints_argv_expect(allocator, scope, argv, 0, AWS_ENDPOINTS_VALUE_BOOLEAN, &argv_value_1) ||
+        aws_endpoints_argv_expect(allocator, scope, argv, 1, AWS_ENDPOINTS_VALUE_BOOLEAN, &argv_value_2)) {
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE, "Failed to resolve booleanEquals.");
+        result = aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_FAILED);
+        goto on_error;
```
TingDaoK, line 211: trivial: ~on_error~ -> on_done

TingDaoK (approved):
fix & ship

source/endpoints_types_impl.c
```diff
@@ -0,0 +1,257 @@
...
+        aws_string_destroy(aws_endpoints_value->v.string.string);
+    }
+
+    if (aws_endpoints_value->type == AWS_ENDPOINTS_VALUE_OBJECT) {
+        aws_string_destroy(aws_endpoints_value->v.object.string);
+    }
+
+    if (aws_endpoints_value->type == AWS_ENDPOINTS_VALUE_ARRAY) {
+        aws_array_list_deep_clean_up(&aws_endpoints_value->v.array, aws_endpoints_value_clean_up_cb);
+    }
```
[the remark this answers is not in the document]
TingDaoK: At least, let's zero out the structure at end, in case it will be reused after clean up.

source/endpoints_rule_engine.c
```diff
@@ -0,0 +1,1081 @@
...
+        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE, "Failed to resolve argv.");
+        goto on_error;
+    }
+
+    if (expected_type != AWS_ENDPOINTS_VALUE_ANY && argv_value.type != expected_type) {
+        AWS_LOGF_ERROR(
+            AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE,
+            "Unexpected arg type actual: %u expected %u.",
+            argv_value.type,
+            expected_type);
```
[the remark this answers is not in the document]
TingDaoK: here as well

PR #24, opened 2023-02-28
Update parsing logic to include sso-session

waahm7:
*Description of changes:*
Changed the profiles map to an array of sections map so that we can handle sso-sessions, which are a different type of section.


source/aws_profile.c
```diff
@@ -30,6 +30,11 @@ struct aws_profile {
...
 };
 
+struct aws_section_collection {
+    struct aws_allocator *allocator;
+    struct aws_hash_table sections;
+};
+
 struct aws_profile_collection {
     struct aws_allocator *allocator;
     enum aws_profile_source_type profile_source;
```
graebm, line 40: I'm making comments about the line below this: struct aws_hash_table profiles; 

1) Change the variable name from ~profiles~ to `sections`. Leaving the name unchanged puts the code into a very confusing state where sometimes "profile" means "profile or session". I get that we can't change the public API, but we can make the internal stuff more understandable going forward.

2) Add a comment documenting what is IN this hashtable. It sucks that our C datastructures are essentially untyped, so comments really help.
```
    /* Map from "profile" or "sso-session" to ???
     * key: struct aws_string*
     * value: struct ???
     */
    struct aws_hash_table sections_by_type
```

OR just have two different hash-tables, one for profiles, one for sso-sessions. It doesn't scale, but it makes sense for now

source/aws_profile.c
```diff
@@ -30,6 +30,11 @@ struct aws_profile {
     bool has_profile_prefix;
 };
 
+struct aws_section_collection {
+    struct aws_allocator *allocator;
+    struct aws_hash_table sections;
+};
```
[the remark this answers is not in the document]
graebm: we talked. needed pointer to allocator later on...

source/aws_profile.c
DmitriyMusatkin: do you really need another hash table? why cant top level just be an array? have a static var 0 refer to profiles, 1 to session tokens and so on for other section types?

include/aws/sdkutils/aws_profile.h
```diff
@@ -99,12 +99,23 @@ const struct aws_profile *aws_profile_collection_get_profile(
...
+ * Returns the number of profiles in a collection
  */
 AWS_SDKUTILS_API
 size_t aws_profile_collection_get_profile_count(const struct aws_profile_collection *profile_collection);
 
+/**
+ * Returns the number of sso-sessions in a collection
+ */
+AWS_SDKUTILS_API
+size_t aws_profile_collection_get_sso_session_count(const struct aws_profile_collection *profile_collection);
```
DmitriyMusatkin, line 117: should we have a generic api that can return map of all values for a given section type and name?

source/aws_profile.c
DmitriyMusatkin: having it as array is arguable: if we want api to always deal with concrete session type (i.e. profiles, session token, etc...) then static array is fine. if we want to support parsing of generic section types then we hash table is probably better

source/aws_profile.c
```diff
@@ -601,28 +607,90 @@ static void s_aws_profile_collection_destroy_internal(struct aws_profile_collect
...
     struct aws_hash_element *element = NULL;
-    aws_hash_table_find(&profile_collection->profiles, profile_name, &element);
+    if (aws_hash_table_find(&profile_collection->profiles, s_profile_token, &element) || element == NULL) {
+        return NULL;
+    }
+    struct aws_section_collection *collection = element->value;
 
-    if (element == NULL) {
+    struct aws_hash_element *profile = NULL;
+    aws_hash_table_find(&collection->sections, profile_name, &profile);
```
DmitriyMusatkin, line 623: nit: hash table find seems inconsistent. first one checks for error, second one ignores it

source/aws_profile.c
```diff
@@ -198,6 +203,7 @@ struct profile_file_parse_context {
     int parse_error;
     int current_line_number;
     bool has_seen_profile;
+    bool is_sso_session;
```
DmitriyMusatkin, line 206: there will be other section types, so you might want to consider basing the logic on whether its profile section or not, rather then whether its not session token section (looks like profile requires separate logic from other section types).

source/aws_profile.c
```diff
@@ -895,6 +991,16 @@ static bool s_parse_profile_declaration(
             return true;
         }
 
+        s_parse_by_character_predicate(&profile_cursor, s_is_whitespace, NULL, 0);
+    } else if (has_sso_session_prefix) {
+        if (context->profile_collection->profile_source == AWS_PST_CREDENTIALS) {
```
DmitriyMusatkin, line 996: if we dont allow profiles or other section types in cred files, do we need to have that check inside of specific section type if

include/aws/sdkutils/aws_profile.h
```diff
@@ -99,12 +99,23 @@ const struct aws_profile *aws_profile_collection_get_profile(
     const struct aws_profile_collection *profile_collection,
     const struct aws_string *profile_name);
 
+AWS_SDKUTILS_API
+const struct aws_profile *aws_profile_collection_get_sso_session(
+    const struct aws_profile_collection *profile_collection,
+    const struct aws_string *profile_name);
```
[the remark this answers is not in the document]
waahm7: Thanks, I have updated it to a generic function.

source/aws_profile.c
waahm7: Thanks, I have updated it to an Array+Enum which is much better in terms of clarity.

include/aws/sdkutils/aws_profile.h
waahm7: Thanks, fixed.

source/aws_profile.c
waahm7: Thanks, fixed by section_type enum.

source/aws_profile.c
waahm7: Not sure how to improve this, as a profile can be default without any prefix. This is only invalid if there was a prefix other than profile, and the source is CREDENTIALS.

source/aws_profile.c
waahm7: Thanks, fixed by refactor.

source/aws_profile.c
```diff
@@ -601,28 +607,90 @@ static void s_aws_profile_collection_destroy_internal(struct aws_profile_collect
...
+    return profile->value;
+}
+
+void aws_section_collection_destroy(void *section_collection) {
+    struct aws_section_collection *collection = section_collection;
+    aws_hash_table_clean_up(&collection->sections);
+    aws_mem_release(collection->allocator, collection);
+}
+
+static struct aws_section_collection *aws_section_collection_new(struct aws_allocator *allocator) {
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed by refactor.

source/aws_profile.c
```diff
@@ -598,20 +604,35 @@ void aws_profile_collection_destroy(struct aws_profile_collection *profile_colle
...
     const struct aws_profile_collection *profile_collection,
     const struct aws_string *profile_name) {
     struct aws_hash_element *element = NULL;
-    aws_hash_table_find(&profile_collection->profiles, profile_name, &element);
-
+    aws_hash_table_find(&profile_collection->sections[AWS_PROFILE_SECTION_TYPE_PROFILE], profile_name, &element);
     if (element == NULL) {
         return NULL;
     }
+    return element->value;
```
graebm, line 624: triviai: just use the new generic function
```suggestion
    return aws_profile_collection_get_section(profile_collection, AWS_PROFILE_SECTION_TYPE_PROFILE, profile_name);
```

source/aws_profile.c
```diff
@@ -620,7 +641,8 @@ static int s_profile_collection_add_profile(
     const struct aws_byte_cursor *profile_name,
     bool has_prefix,
     const struct profile_file_parse_context *context,
-    struct aws_profile **current_profile_out) {
+    struct aws_profile **current_profile_out,
+    const enum aws_profile_section_type section_type) {
```
graebm, line 645: similar to earlier comment, move `section_type` arg so that it comes before `profile_name`

graebm: also, debatable, but does it make to code more clear going forward if we change ~profile_name~ -> `section_name` or `name`?

graebm (approved):
fix & ship

DmitriyMusatkin (approved):

source/aws_profile.c
waahm7: I think it would be better to stick to one name instead of mixing and matching.

PR #30, opened 2023-05-24
update the behavior for profile override

TingDaoK:
*Issue #, if available:*

- We say we do what aws-cli does, [here](https://github.com/awslabs/aws-c-sdkutils/blob/main/include/aws/sdkutils/aws_profile.h#L24-L27)
- But, according to CLI doc [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html):
>   Specifies the name of the AWS CLI profile with the credentials and options to use. This can be the name of a profile stored in a credentials or config file, or the value default to use the default profile.
    If defined, this environment variable overrides the behavior of using the profile named [default] in the configuration file. You can override this environment variable by using the --profile command line parameter.

*Description of changes:*

- Change the behavior to make sure the provided `profile` override the environment variable. 

### DIFF
--- a/source/aws_profile.c
+++ b/source/aws_profile.c
@@ -1542,16 +1542,18 @@ AWS_STATIC_STRING_FROM_LITERAL(s_default_profile_env_variable_name, "AWS_PROFILE
 
 struct aws_string *aws_get_profile_name(struct aws_allocator *allocator, const struct aws_byte_cursor *override_name) {
     /**
-     * The default profile must be used if the user does NOT specify a profile.
-     * the default profile can be overridden using the AWS_PROFILE environment variable.
-     */
+     * Profile name is resolved in the following order.
+     * 1. If the override_path variable is provided.
+     * 2. Check `AWS_PROFILE` environment variable and use the value if it is not empty.
+     * 3. Use "default". */
     struct aws_string *profile_name = NULL;
     if (override_name != NULL && override_name->ptr != NULL) {
         profile_name = aws_string_new_from_array(allocator, override_name->ptr, override_name->len);
     } else {
-        /* Use the default profile. Override the default profile with AWS_PROFILE environment variable if exist */
-        if (aws_get_environment_value(allocator, s_default_profile_env_variable_name, &profile_name) ||
-            profile_name == NULL) {
+        /* Try to fetch profile from AWS_PROFILE environment variable */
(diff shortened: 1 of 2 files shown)


source/aws_profile.c
```diff
@@ -1541,14 +1541,17 @@ struct aws_string *aws_get_config_file_path(
...
-        profile_name == NULL) {
-        if (override_name != NULL && override_name->ptr != NULL) {
-            profile_name = aws_string_new_from_array(allocator, override_name->ptr, override_name->len);
-        } else {
+    if (override_name != NULL && override_name->ptr != NULL) {
+        profile_name = aws_string_new_from_array(allocator, override_name->ptr, override_name->len);
+    } else {
+        /* Use the default profile. Override the default profile with AWS_PROFILE environment variable if exist */
+        if (aws_get_environment_value(allocator, s_default_profile_env_variable_name, &profile_name) ||
+            profile_name == NULL) {
```
waahm7, line 1554: Existing logic but I think it would be more clear to split this. Something like 
```suggestion
        /* Try to fetch profile from AWS_PROFILE environment variable */
        aws_get_environment_value(allocator, s_default_profile_env_variable_name, &profile_name)
        /* Use default profile if it doesn't exist. */
        if (profile_name == NULL) {
```

source/aws_profile.c
```diff
@@ -1541,14 +1541,17 @@ struct aws_string *aws_get_config_file_path(
 AWS_STATIC_STRING_FROM_LITERAL(s_default_profile_env_variable_name, "AWS_PROFILE");
 
 struct aws_string *aws_get_profile_name(struct aws_allocator *allocator, const struct aws_byte_cursor *override_name) {
-
+    /**
+     * The default profile must be used if the user does NOT specify a profile.
+     * the default profile can be overridden using the AWS_PROFILE environment variable.
+     */
```
waahm7, line 1547: We should update the docs in header file to list the order the resolution.

waahm7 (approved):
fix & ship.

source/aws_profile.c
waahm7: I think it would be more clear to list it something like
```
/**
* Profile name is resolved in the following order.
* 1. If the override_path variable is provided, use that.
* 2. Check `AWS_PROFILE` and use that if it is not empty.
* 3. Use the `default` profile name.
*/
```

PR #31, opened 2023-07-08
aws_profile: reduce logging verbosity of the ~/.aws/profile parser

grrtrr:
The profile parser is used frequently and in many places. Its logging level is very verbose. Even when it is clear that `~/.aws/profile` does not exist (e.g. on kubernetes nodes using STS), it repeatedly logs entries like these:
```console
[ERROR] 2022-03-14 22:05:06.078 file-utils [139977917552384] static: Failed to open file /root/.aws/config with errno 2
[ERROR] 2022-03-14 22:05:06.088 file-utils [139977917552384] static: Failed to open file /root/.aws/credentials with errno 2
```
We have found that the verbose logs confuse users in production, causing unnecessary tickets to be raised.

Reduce the verbosity of frequently used logging messages to `DEBUG`; to allow seeing the messages when troubleshooting.

### DIFF
--- a/source/aws_profile.c
+++ b/source/aws_profile.c
@@ -346,7 +346,7 @@ static int s_profile_property_add_sub_property(
     int was_present = 0;
     aws_hash_table_remove(&property->sub_properties, key_string, NULL, &was_present);
     if (was_present) {
-        AWS_LOGF_WARN(
+        AWS_LOGF_DEBUG(
             AWS_LS_SDKUTILS_PROFILE,
             "subproperty \"%s\" of property \"%s\" had value overridden with new value",
             key_string->bytes,
@@ -385,7 +385,7 @@ static int s_profile_property_merge(struct aws_profile_property *dest, const str
         }
 
         if (dest->value) {
-            AWS_LOGF_WARN(
+            AWS_LOGF_DEBUG(
                 AWS_LS_SDKUTILS_PROFILE,
                 "property \"%s\" has value \"%s\" replaced during merge",
                 dest->name->bytes,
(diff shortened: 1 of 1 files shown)


source/aws_profile.c
```diff
@@ -1039,15 +1039,15 @@ static bool s_parse_property_continuation(
      * A continuation without a current property is bad
      */
     if (context->current_profile == NULL || context->current_property == NULL) {
-        AWS_LOGF_WARN(AWS_LS_SDKUTILS_PROFILE, "Property continuation seen outside of a current property");
+        AWS_LOGF_DEBUG(AWS_LS_SDKUTILS_PROFILE, "Property continuation seen outside of a current property");
```
waahm7, line 1042: This is a fatal error and it should not occur frequently unless your profile is messed up. Maybe: 
```suggestion
        AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PROFILE, "Property continuation seen outside of a current property");
```

source/aws_profile.c
```diff
@@ -1155,7 +1155,7 @@ static bool s_parse_property(const struct aws_byte_cursor *line_cursor, struct p
         context->current_property =
             s_profile_add_property(context->current_profile, &trimmed_key_cursor, &property_cursor);
         if (context->current_property == NULL) {
-            AWS_LOGF_ERROR(
+            AWS_LOGF_DEBUG(
```
waahm7, line 1158: Fatal error, I suggest we keep it as error.

source/aws_profile.c
```diff
@@ -1214,7 +1214,7 @@ static void s_parse_and_apply_line_to_profile_collection(
         return;
     }
 
-    AWS_LOGF_WARN(AWS_LS_SDKUTILS_PROFILE, "Unidentifiable line type encountered while parsing profile file");
+    AWS_LOGF_DEBUG(AWS_LS_SDKUTILS_PROFILE, "Unidentifiable line type encountered while parsing profile file");
```
waahm7, line 1217: Again fatal error and should not occur frequently. 
```suggestion
    AWS_LOGF_ERROR(AWS_LS_SDKUTILS_PROFILE, "Unidentifiable line type encountered while parsing profile file");
```

waahm7 (changes requested):

waahm7 (approved):
Thank you for creating the PR and incorporating the changes.

PR #46, opened 2024-10-11
Remove arg with cxx keyword from header.

jeffdav:
*Issue #, if available:*
When building a Swift module with cxx-interop enabled, this module gets imported by clang as a C++ header.  Since `template` is a C++ keyword, it causes problems.

*Description of changes:*
Just comment out the argument name in the header.  C only requires the type.

### DIFF
--- a/include/aws/sdkutils/private/endpoints_util.h
+++ b/include/aws/sdkutils/private/endpoints_util.h
@@ -74,7 +74,7 @@ AWS_SDKUTILS_API void aws_array_list_deep_clean_up(
 
 /* Function that resolves template. */
 typedef int(aws_endpoints_template_resolve_fn)(
-    struct aws_byte_cursor template,
+    struct aws_byte_cursor /* template */,
     void *user_data,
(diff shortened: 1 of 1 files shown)


bretambrose:
Without knowing the details, my instinct is that " this module gets imported by clang as a C++ header" is the problem that needs to be fixed.

jeffdav:
Heh, I don't disagree. :)

However, what's happening is we're using `aws-sdk-swift`, which depends on these modules via `aws-crt-swift`.  However a lot of our project, including the bits that use `aws-sdk-swift` are built with [cxx-interoperability-mode=default](https://www.swift.org/documentation/cxx-interop/project-build-setup/#mixing-swift-and-c-using-other-build-systems) for other reasons and so I'm kind of stuck.

subdiox:
I think that is because `template` is a reserved keyword. I put up a similar PR https://github.com/awslabs/aws-c-sdkutils/pull/48 unintentionally, but I hope either of them is merged.

PR #51, opened 2025-01-06
Update Config File Parsing Logic to Parse Services Section

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/308

*Description of changes:*
- Config files have a new valid section named "services"; update our parser to handle that section. See: https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html
- Refactor tests to reduce duplication for different types of sections.


source/aws_profile.c
```diff
@@ -862,6 +863,31 @@ static struct aws_byte_cursor s_trim_trailing_whitespace_comment(const struct aw
     return trimmed;
 }
 
+static int s_parse_section_type_prefix(
+    struct aws_byte_cursor *profile_cursor,
+    enum aws_profile_section_type *out_section_type) {
+    /* check profile prefix */
+    if (s_parse_by_token(profile_cursor, s_profile_token, NULL) &&
+        s_parse_by_character_predicate(profile_cursor, s_is_whitespace, NULL, 1)) {
```
graebm, line 871: bug in theory: if `s_parse_by_token()` returns true but `s_parse_by_character_predicate()` returns false, then `profile_cursor` is unexpectedly advanced, and the next `if (s_parse_by_token(....` block will do unexpected things.

So like it starts with invalid nonsense like "profilesso-session", then the first if-block in this function will advance the cursor to be "sso-session" and then the second if-block would pass, but like ... neither should pass

source/aws_profile.c
```diff
@@ -862,6 +863,31 @@ static struct aws_byte_cursor s_trim_trailing_whitespace_comment(const struct aw
...
+        *out_section_type = AWS_PROFILE_SECTION_TYPE_PROFILE;
+        return AWS_OP_SUCCESS;
+    }
+    /* check sso-session prefix */
+    if (s_parse_by_token(profile_cursor, s_sso_session_token, NULL) &&
+        s_parse_by_character_predicate(profile_cursor, s_is_whitespace, NULL, 1)) {
```
graebm, line 877: s_is_whitespace() also allows '\n' which is .... maybe not what we want here
```
[profile
bozo]
```
shouldn't pass, right?

source/aws_profile.c
waahm7: Thanks, that's a great catch. I have fixed it and added a test.

source/aws_profile.c
waahm7: Thanks, this is not a problem since we split the config file by \n first https://github.com/awslabs/aws-c-sdkutils/blob/db8625aea8c38861d1005fe7869cc63909ba454b/source/aws_profile.c#L1285. So
```
[profile
bozo]
```
is invalid since it will be treated as 2 individual lines.

graebm (approved):

PR #53, opened 2025-05-14
Fix double free on malformed rulesets

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Note: its invalid for ruleset to have a required param without a default (this is enforced by upstream validation). resolver also sanity checks that assertion, but in certain cases of failure it can endup freeing some allocated state twice. So lets avoid doing that. 


waahm7 (approved):

source/partitions.c
```diff
@@ -201,6 +203,7 @@ static int s_on_partition_element(
     return AWS_OP_SUCCESS;
 
 on_error:
+    aws_partition_info_destroy(partition_info);
```
TingDaoK, line 206: at line 200, we may destroy the partition_info that already put into the hashtable from line 185, right?

TingDaoK (approved):

PR #58, opened 2025-08-07
Fix warnings found by the Undefined Behavior Sanitizer

bjosv:
Fixes runtime error warnings found when running the tests with the undefined behavior sanitizer enabled during build.

```
aws-c-common/source/ref_count.c:29:9: runtime error: call to function s_aws_profile_collection_destroy_internal through pointer to incorrect function type 'void (*)(void *)'
aws-c-sdkutils/source/endpoints_util.c:50:24: runtime error: addition of unsigned offset to 0x566df83f0480 overflowed to 0x566df83f047f

```
- Update callback functions to use correct signature, see [aws_simple_completion_callback](https://github.com/awslabs/aws-c-common/blob/7b24a8a4342df68b3e461353cb8f3aa95b070d59/include/aws/common/shutdown_types.h#L13).
- Add check of `label.len` so we don't overflow/wrap at line 50 `return aws_isalnum(label.ptr[label.len - 1]);`

Since the CI-build is not using `-fno-sanitize-recover=all` the warnings are recoverable and will not fail the CI build.
Warnings can be found in test logs (`grep "runtime error" Testing/Temporary/LastTest.log`)


sfod:
Thank you for your contribution. Your work will be merged in https://github.com/awslabs/aws-c-sdkutils/pull/75
Closing this PR.

PR #60, opened 2026-02-25
Ingest BDD endpoints

azkrishpy:
*Issue #, if available:*

*Description of changes:*


include/aws/sdkutils/endpoints_bdd_engine.h
```diff
@@ -0,0 +1,68 @@
...
+AWS_PUSH_SANE_WARNING_LEVEL
+
+struct aws_endpoints_bdd_engine;
+struct aws_partitions_config;
+struct aws_endpoints_request_context;
+struct aws_endpoints_resolved_endpoint;
+
+AWS_EXTERN_C_BEGIN
+
+/**
```
azkrishpy, line 21: Fix comment.

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,371 @@
+#!/usr/bin/env python3
```
azkrishpy, line 1: License

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,986 @@
...
+    if (cursor->len < blob_size) {
+        return AWS_OP_ERR;
+    }
+
+    *out_blob = aws_byte_cursor_from_array(cursor->ptr, blob_size);
+    aws_byte_cursor_advance(cursor, blob_size);
+    return AWS_OP_SUCCESS;
+}
+
+static int s_read_string_ref(
```
azkrishpy: is this necessary? even if necessary, can't we set it to the *cursor directly?

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,374 @@
...
+        buf += struct.pack("<BH", 6, len(val))
+        for item in val:
+            encode_value(buf, item, strings)
+    elif isinstance(val, dict):
+        buf += struct.pack("<BH", 7, len(val))
+        for k, v in val.items():
+            write_string_ref(buf, *strings.add(k))
+            encode_value(buf, v, strings)
+    else:
+        buf += struct.pack("<B", 0)
```
azkrishpy: Should we error out instead?

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,986 @@
...
+                goto error;
+            }
+            AWS_LOGF_DEBUG(
+                AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE, "Result %d has %d properties", (int)i, (int)property_count);
+
+            if (aws_byte_buf_init(&result.data.endpoint.properties, allocator, 256)) {
+                goto error;
+            }
+
+            /* Encode properties as JSON object */
```
azkrishpy: why JSON? JSON bad.

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def __init__(self):
+        self._blob = bytearray()
+        self._index = {}  # str -> (offset, length)
+
+    def add(self, s):
+        if not s:
+            return (0, 0)
+        if s not in self._index:
+            encoded = s.encode("utf-8")
```
DmitriyMusatkin: dont think we support unicode currently. lets just error out for now

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,382 @@
...
+            offset = len(self._blob)
+            if offset > 0xFFFF:
```
DmitriyMusatkin, line 126: how close is s3 ruleset to this limit?

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,382 @@
...
+                raise ValueError(f"String blob offset {offset} exceeds uint16 max")
+            if len(encoded) > 0xFFFF:
+                raise ValueError(f"String length {len(encoded)} exceeds uint16 max")
+            self._blob.extend(encoded)
```
DmitriyMusatkin: lets also insert some special chars between strings for easier debugging and some potential error checking

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,382 @@
...
+            _collect_value_strings(v, strings)
+    elif isinstance(val, list):
+        for v in val:
+            _collect_value_strings(v, strings)
+
+
+def encode_parameters(buf, data, strings):
+    """Emit parameter count (uint16) then each parameter record."""
+    params = data.get("parameters", {})
+    buf += struct.pack("<H", len(params))
```
DmitriyMusatkin: can you document exact format somewhere? maybe at the top of this file?

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,382 @@
...
+    for name, p in params.items():
+        ptype = p.get("type", "string").lower()
+        opcode = 0x01 if ptype == "string" else 0x02
+        buf += struct.pack("<B", opcode)
+        write_string_ref(buf, *strings.add(name))
```
DmitriyMusatkin: so it does a prepass and then goes over all string again as it inserts them? not sure i follow why prepass is needed?

compiler/bdd-bytecoder.py
```diff
@@ -0,0 +1,382 @@
...
+        dv = p.get("default")
+        has_default = dv is not None
+        buf += struct.pack("<B", 1 if has_default else 0)
+        if has_default:
+            if opcode == 0x01:
```
DmitriyMusatkin: default by definition is boolean, bo?

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -139,6 +140,14 @@ struct aws_endpoints_function {
     struct aws_array_list argv;
 };
 
+struct aws_endpoints_expr; /* Forward declaration */
+
+struct aws_endpoints_kv_pair {
```
DmitriyMusatkin, line 145: which construct does this map to in bdd?

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -328,4 +338,50 @@ int aws_endpoints_path_through_object(
...
+
+    struct aws_hash_table parameters;
+
+    struct aws_array_list conditions;
+
+    struct aws_array_list results;
+
+    int32_t root_ref;
+    uint32_t node_count;
+    struct aws_endpoints_bdd_node *nodes;
```
DmitriyMusatkin: should this be wrapped in array_list as well?

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,838 @@
...
+static void s_on_condition_array_element_clean_up(void *element) {
+    struct aws_endpoints_condition *condition = element;
+    aws_endpoints_condition_clean_up(condition);
+}
+
+static int s_read_u16(struct aws_byte_cursor *cursor, uint16_t *out) {
+    if (cursor->len < 2) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+    *out = (uint16_t)cursor->ptr[0] | ((uint16_t)cursor->ptr[1] << 8);
```
DmitriyMusatkin: this will break on big endian. dont we have helpers for this in common?

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,838 @@
...
+        if (opcode != 0x01 && opcode != 0x02) {
+            goto error;
+        }
+
+        struct aws_endpoints_parameter *param = aws_mem_calloc(allocator, 1, sizeof(struct aws_endpoints_parameter));
+        if (!param) {
+            goto error;
+        }
+        param->allocator = allocator;
+        param->type = (opcode == 0x01) ? AWS_ENDPOINTS_PARAMETER_STRING : AWS_ENDPOINTS_PARAMETER_BOOLEAN;
```
DmitriyMusatkin: just have opcode value match enum value to avoid the check or some sort of static map? param can also be string array if i remember correctly

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,838 @@
...
+            goto error;
+        }
+
+        /* Construct FUNCTION expression from fn_name and argc */
+        cond.expr.type = AWS_ENDPOINTS_EXPR_FUNCTION;
+        cond.expr.e.function.fn = AWS_ENDPOINTS_FN_LAST;
+
+        /* Resolve function name to enum */
+        uint64_t hash = aws_hash_byte_cursor_ptr(&fn_name);
+        for (int idx = AWS_ENDPOINTS_FN_FIRST; idx < AWS_ENDPOINTS_FN_LAST; ++idx) {
```
DmitriyMusatkin: consider helpers here instead of repeating the code inline

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,838 @@
...
+            }
+            if (aws_array_list_push_back(&cond.expr.e.function.argv, &arg)) {
+                aws_endpoints_expr_clean_up(&arg);
+                aws_array_list_deep_clean_up(&cond.expr.e.function.argv, s_on_expr_array_element_clean_up);
+                goto error;
+            }
+        }
+
+        uint8_t has_assign;
+        if (cursor->len < 1) {
```
DmitriyMusatkin: maybe consider just using advance? it returns the advanced part and then you can check whether it matches the expected size?
also consider breaking up functions into helpers and just cleaning up once in the on_error handler. current approach has a potential to forget to clean something

compiler/bdd-bytecoder.py
azkrishpy: I think after adding all possible strings it was around 200? should be fine I think.

compiler/bdd-bytecoder.py
azkrishpy: sounds good

compiler/bdd-bytecoder.py
azkrishpy: good callout. will do!

compiler/bdd-bytecoder.py
azkrishpy: this method adds strings to the _index dict if the string is not already present. Do you mean we don't encode it at all just store as strings?

include/aws/sdkutils/private/endpoints_types_impl.h
azkrishpy: yes, dumb mistake.

include/aws/sdkutils/private/endpoints_types_impl.h
azkrishpy: Any object (key value pair in bdd). I think previously we used to treat it as a json that needs to be resolved at runtime. this simplifies it as something we already have structured.

source/endpoints_bdd_loader.c
azkrishpy: by default all integers encoded into bytecode are little endian. we need to support both while creating bytecode as well or just read things differently based on what machine type it is?

compiler/bdd-bytecoder.py
azkrishpy: true, thought there might be string defaults. will fix this.

compiler/bdd-bytecoder.py
azkrishpy: great callout. the prepass was because the we were storing the blob first before encoding the rest. But now that I think about it, we can move the string blob to the end and we would not have to prepass. That said, is processing time a real enough concern for us to change this? because this would mean we have to read the string blob first from the bytecode before adapting to our structs which could complicate things a bit on native.

compiler/bdd-bytecoder.py
DmitriyMusatkin: yes. we would not be able to parse it on c side if string contains unicode

compiler/bdd-bytecoder.py
DmitriyMusatkin: you dont have to dump all sections to final blob as you parse the file. construct all individual segments in mem and the write them in order to final output. 
tradeoff here is do 2 passes, but construct linearly. but potentially risk minor differences between passes causing really hard to debug issues
or do one pass and waste some mem

compiler/bdd-bytecoder.py
DmitriyMusatkin: has_default will always be boolean? default value itself can be string or bool

include/aws/sdkutils/private/endpoints_types_impl.h
DmitriyMusatkin: oh is this the properties json in endpoint result? does bdd flatten it? if not how does this deal with nested json structures?

source/endpoints_bdd_loader.c
DmitriyMusatkin: its fine if its encoded in little endian or big endian in bytecode, but reader needs to be aware of that. dont think the above will work as expected on the big endian machine

DmitriyMusatkin: we have helpers in common that can handle endianess and if we dont for specific case i would start by adding them to common

compiler/bdd-bytecoder.py
azkrishpy: yep, that is better. Will change to do a single pass instead.

compiler/bdd-bytecoder.py
azkrishpy: makes sense, will store it as just strings as is

compiler/bdd-bytecoder.py
azkrishpy: we should error out since getting a value that is not any of these types in unexpected

source/endpoints_bdd_loader.c
azkrishpy: makes sense, will make the change

source/endpoints_bdd_loader.c
azkrishpy: should be fixed now, thanks!

source/endpoints_bdd_loader.c
azkrishpy: I will break up functions that use helpers.

source/endpoints_bdd_loader.c
azkrishpy: interesting! Will check out c-common for the helper.

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,724 @@
...
+
+static void s_on_condition_array_element_clean_up(void *element) {
+    struct aws_endpoints_condition *condition = element;
+    aws_endpoints_condition_clean_up(condition);
+}
+
+/*
+ * Helper: read a big-endian i32 from cursor.
+ */
+static int s_read_i32(struct aws_byte_cursor *cursor, int32_t *out) {
```
DmitriyMusatkin: can you do a quick pr to just have this function in common?

source/endpoints_bdd_loader.c
```diff
@@ -0,0 +1,724 @@
...
+    aws_byte_cursor_advance(cursor, blob_size);
+    return AWS_OP_SUCCESS;
+}
+
+static int s_read_string_ref(struct aws_byte_cursor *cursor, struct aws_byte_cursor blob, struct aws_byte_cursor *out) {
+    uint16_t offset, length;
+    if (!aws_byte_cursor_read_be16(cursor, &offset) || !aws_byte_cursor_read_be16(cursor, &length)) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+    if ((size_t)offset + length > blob.len) {
```
DmitriyMusatkin, line 81: nit: does this need a cast? uint16_t should always be smaller than size_t

source/endpoints_bdd_loader.c
azkrishpy: addressed in https://github.com/awslabs/aws-c-common/pull/1239 and switched to use the new helper aws_byte_cursor_read_be_i32()

source/endpoints_bdd_loader.c
azkrishpy: the total could exceed the uint16_t so leaving this as is.

DmitriyMusatkin (approved):

PR #62, opened 2026-05-23
BDD engine implementation

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Add engine that can evaluate bdd endpoint resolution requests. Builds on top of the previous bdd loader pr and basically takes the bdd internal representation + context and returns resolved endpoint. 
Engine and tests are added in this pr. Perf testing scaffolding will be added in the follow up.


source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,328 @@
...
+on_error:
+    aws_endpoints_scope_value_destroy(scope_value);
+    /* Only cleanup value if mem ownership was not transferred to scope value. */
+    if (scope_value == NULL) {
+        aws_endpoints_value_clean_up(&val);
+    }
+
+    *out_is_truthy = false;
+    return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_FAILED);
+}
```
azkrishpy, line 185: seems identical to previous engine's implementation.. should we consider reusing?

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,328 @@
...
+
+    int32_t result_idx = (current_ref - s_result_bound);
+    struct aws_endpoints_bdd_result eval_result;
+    if (aws_array_list_get_at(&engine->results, &eval_result, result_idx)) {
+        result = aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_FAILED);
+        goto on_done;
+    }
+
+    if (eval_result.type == AWS_ENDPOINTS_RESOLVED_ENDPOINT) {
+        struct aws_endpoints_resolved_endpoint *endpoint = aws_endpoints_resolved_endpoint_new(engine->allocator);
```
azkrishpy, line 264: should we be releasing it when we error and go to on_done?

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,328 @@
...
+            goto on_done;
+        }
+
+        bool is_thruthy = false;
+        if (s_resolve_one_condition(engine->allocator, current_condition, &state, &is_thruthy)) {
+            result = aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_FAILED);
+            goto on_done;
+        }
+
+        if (is_complement ^ is_thruthy) {
```
azkrishpy:
```suggestion
        bool is_truthy = false;
        if (s_resolve_one_condition(engine->allocator, current_condition, &state, &is_truthy)) {
            result = aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_FAILED);
            goto on_done;
        }

        if (is_complement ^ is_truthy) {
```

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,328 @@
...
+
+    struct aws_endpoints_bdd_engine *engine;
+};
+
+static void s_scope_value_destroy_cb(void *data) {
+    struct aws_endpoints_scope_value *value = data;
+    aws_endpoints_scope_value_destroy(value);
+}
+
+/* TODO: does it need to be deep copy */
```
azkrishpy: seems like we don't refcount context, maybe we are worried about early release? previous version also does a deep copy.

azkrishpy:
we should add update description with what the PR accomplishes and what is improved

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,328 @@
...
+        scope_value->value = val;
+
+        int was_created = 1;
+        if (aws_hash_table_put(&state->scope.values, &scope_value->name.cur, scope_value, &was_created)) {
+            AWS_LOGF_ERROR(AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE, "Failed to set assigned variable.");
+            goto on_error;
+        }
+
+        /* Shadowing existing values is prohibited. */
+        if (!was_created) {
```
azkrishpy: question: what happens to the scope.values table when we put a key value pair and then destroy the key on_error?

source/endpoints_bdd_engine.c
DmitriyMusatkin: this has been completely changed to allow for value reg premapping

tests/endpoint_bdd_tests.c
```diff
@@ -0,0 +1,196 @@
...
+
+    struct aws_byte_buf partitions_buf;
+    ASSERT_SUCCESS(aws_byte_buf_init_from_file(&partitions_buf, allocator, "sample_partitions.json"));
+    struct aws_byte_cursor partitions_json = aws_byte_cursor_from_buf(&partitions_buf);
+
+    struct aws_partitions_config *partitions = aws_partitions_config_new_from_string(allocator, partitions_json);
+    ASSERT_NOT_NULL(partitions);
+
+    struct aws_endpoints_bdd_engine *engine =
+        aws_endpoints_bdd_engine_new_from_bytecode(allocator, aws_byte_cursor_from_buf(&bytecode), partitions);
```
TingDaoK, line 28: we are not copying the bytecode and it's not refcounted

We need to document it out that bytecode should out live the bdd engine. Or should we consider to deep copy the necessary part?

source/endpoints_bdd_loader.c
```diff
@@ -653,74 +676,68 @@ struct aws_endpoints_bdd_engine *aws_endpoints_bdd_engine_new_from_bytecode(
...
+        struct aws_endpoints_parameter *value = (struct aws_endpoints_parameter *)iter.element.value;
+
+        struct aws_hash_element *element = NULL;
+        aws_hash_table_find(&engine->register_map, &value->name, &element);
+
+        if (element == NULL) {
+            aws_hash_table_put(&engine->register_map, &value->name, (void *)param_count, NULL);
+            param_count++;
+        }
+    }
```
TingDaoK: we should add comments around this `register_map`, it's fairly hard to understand what is is and how it's used.
looks like we are iterating through the params table and create another table from the name to index? 

Why not add it as part of `s_load_parameters`?

Also, in this case, why do we need two hash table? Will it be simpler to have a hash table from name to index and then one array directly?

source/endpoints_bdd_loader.c
```diff
@@ -753,6 +766,7 @@ static void s_endpoints_bdd_engine_destroy(void *data) {
     }
 
     aws_hash_table_clean_up(&engine->parameters);
+    aws_hash_table_clean_up(&engine->register_map);
     aws_array_list_deep_clean_up(&engine->conditions, s_on_condition_array_element_clean_up);
     aws_mem_release(engine->allocator, engine->conditions_ptr);
```
TingDaoK: not part of this PR, but we don't need this deep clean up of the conditions. All it does is zero out each condition in the array and we release the full array one line afterward.

TingDaoK: trivial: ~conditions_ptr~ -> conditions_array_ptr

source/endpoints_bdd_loader.c
```diff
@@ -653,74 +676,68 @@ struct aws_endpoints_bdd_engine *aws_endpoints_bdd_engine_new_from_bytecode(
...
+         aws_hash_iter_next(&iter)) {
+
+        struct aws_endpoints_parameter *value = (struct aws_endpoints_parameter *)iter.element.value;
+
+        struct aws_hash_element *element = NULL;
+        aws_hash_table_find(&engine->register_map, &value->name, &element);
+
+        if (element == NULL) {
+            aws_hash_table_put(&engine->register_map, &value->name, (void *)param_count, NULL);
+            param_count++;
```
TingDaoK: looks like the `register_map` has a limit of `s_max_regs`, which is 128, but this param_count is a uint16_t, the value we add to the `engine->register_map` needs to be protected against the `s_max_regs`?

source/endpoints_bdd_loader.c
```diff
@@ -653,74 +676,68 @@ struct aws_endpoints_bdd_engine *aws_endpoints_bdd_engine_new_from_bytecode(
...
     }
-    // clock_t end_magic = clock();
-    // double time_taken_magic = (((double)(end_magic - begin_magic)) / 1000);
-    // AWS_LOGF_INFO(AWS_LS_SDKUTILS_ENDPOINTS_PARSING, "BDD magic in(ms): %f", time_taken_magic);
 
-    // clock_t begin_string = clock();
+    if (aws_hash_table_init(
+            &engine->register_map,
+            allocator,
+            s_max_regs,
```
TingDaoK: this `s_max_regs` is just initial cap, the hash table will grow automatically.

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/sdkutils/endpoints_bdd_engine.h>
+#include <aws/sdkutils/private/endpoints_types_impl.h>
+
+struct aws_bdd_scope {
+    struct aws_endpoints_bdd_engine *engine;
+
+    struct aws_endpoints_scope_value values[s_max_regs];
```
TingDaoK: oof, i feel like we will get loooot of ai security ticket for this fixed array

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+
+        struct aws_hash_element *element = NULL;
+        aws_hash_table_find(&state->engine->register_map, &key, &element);
+
+        uint16_t idx = 0;
+        if (element != NULL) {
+            idx = (uint16_t)(uintptr_t)element->value;
+        }
+
+        if (state->scope_impl.values[idx].value.type == AWS_ENDPOINTS_VALUE_ANY) {
```
TingDaoK: the idex can be more than `s_max_regs` and result in out of bound access.

TingDaoK: Also, `AWS_ENDPOINTS_VALUE_ANY` naming is confusing, should it be something like NOT_VALID, NOT_INITED?

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+
+    if (s_copy_context_to_state(context, state)) {
+        return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_INIT_FAILED);
+    }
+
+    /* Add defaults to the top level scope. */
+    for (struct aws_hash_iter iter = aws_hash_iter_begin(&engine->parameters); !aws_hash_iter_done(&iter);
+         aws_hash_iter_next(&iter)) {
+        const struct aws_byte_cursor key = *(const struct aws_byte_cursor *)iter.element.key;
+        struct aws_endpoints_parameter *value = (struct aws_endpoints_parameter *)iter.element.value;
```
TingDaoK: trivial: there are so many different `values` in this function... maybe renaming some of those to increase readability.

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -442,6 +456,10 @@ struct aws_endpoints_bdd_result {
     } data;
 };
 
+enum {
+    s_max_regs = 128, /* max regs during eval */
+};
```
TingDaoK: from my reading, we have each scope owning an fix size array of values that index is the hash value from `register_map` as a frequently used values for reference.

Why not the engine itself maintain the array?

It seems to be fragile that engain maintains the map, but the real value map refer to is maintained somewhere else.

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+
+        struct aws_endpoints_scope_value *context_value = (struct aws_endpoints_scope_value *)iter.element.value;
+        struct aws_hash_element *element = NULL;
+        aws_hash_table_find(&state->engine->register_map, &context_value->name, &element);
+
+        size_t idx = 0;
+        if (element != NULL) {
+            idx = (size_t)element->value;
+        } else {
+            idx = aws_hash_table_get_entry_count(&state->engine->register_map) + 1;
```
TingDaoK: this may overflow, since the register map can grow.

TingDaoK: And, we are not adding this context to the map?

include/aws/sdkutils/private/endpoints_types_impl.h
TingDaoK: Is this number an enforcement from bdd spec?

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+            continue;
+        }
+
+        struct aws_hash_element *element = NULL;
+        aws_hash_table_find(&state->engine->register_map, &key, &element);
+
+        uint16_t idx = 0;
+        if (element != NULL) {
+            idx = (uint16_t)(uintptr_t)element->value;
+        }
```
TingDaoK: else? we should at least error out? Or it will just continue with idex 0

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+
+static void s_state_clean_up(struct aws_endpoints_bdd_engine_state *state) {
+    AWS_PRECONDITION(state);
+
+    for (size_t i = 0; i < s_max_regs; ++i) {
+        aws_endpoints_value_clean_up(&state->scope_impl.values[i].value);
+    }
+}
+
+static int32_t s_result_bound = 100000000;
```
TingDaoK: trivial: mark it const.

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+
+    if (ref.bdd_ref_idx == 0) {
+        struct aws_hash_element *element = NULL;
+        aws_hash_table_find(&bdd_scope->engine->register_map, &ref.name, &element);
+        if (element == NULL) {
+            return NULL;
+        }
+
+        size_t reg_idx = (size_t)element->value;
+        ret = &((struct aws_bdd_scope *)scope_impl)->values[reg_idx];
```
azkrishpy: we should add the value we found to bdd_ref_idx we have now found so that we do not keep doing querying the hash table.

source/endpoints_bdd_engine.c
azkrishpy: this logic is hard to understand.

source/endpoints_bdd_loader.c
TingDaoK: and looks like everywhere else will only add new field to the table, not override existing key. 
So, I think we can safely initialize the register map with all params and store params as array.

And if any params is not in the map, we should error out.

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+
+    *out_is_truthy = aws_endpoints_is_value_truthy(&val);
+
+    if (condition->assign.len > 0) {
+        struct aws_bdd_scope *scope_impl = &state->scope_impl;
+        struct aws_endpoints_scope_value *scope_value = &scope_impl->values[condition->assign_idx];
+
+        scope_value->allocator = NULL;
+        scope_value->name = aws_endpoints_non_owning_cursor_create(condition->assign);
+        scope_value->value = val;
```
TingDaoK, line 174: any chance that the condition assign can override an existing value?

source/endpoints_bdd_engine.c
```diff
@@ -0,0 +1,322 @@
...
+    aws_array_list_init_static_from_initialized(
+        &state->scope.expr_index, engine->expr_ptr, engine->expr_len, sizeof(struct aws_endpoints_expr));
+
+    if (s_copy_context_to_state(context, state)) {
+        return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_INIT_FAILED);
+    }
+
+    /* Add defaults to the top level scope. */
+    for (struct aws_hash_iter iter = aws_hash_iter_begin(&engine->parameters); !aws_hash_iter_done(&iter);
+         aws_hash_iter_next(&iter)) {
```
TingDaoK: instead of initailize the default state each time, can we instead store the default values with the engine, and skip this part for every init_state?

source/endpoints_bdd_engine.c
azkrishpy: no

source/endpoints_bdd_engine.c
azkrishpy: AWS_ENDPOINTS_VALUE_ANY represents any of the possible endpoint value types.. as in NOT NONE. I think the naming convention here is fine.

include/aws/sdkutils/private/endpoints_types_impl.h
azkrishpy: no enforced number from spec but should not exceed 128 (from what i read: 17 parameters, 20 condition assigns)

source/endpoints_bdd_engine.c
azkrishpy: this is not worth optimizing as the array which we would copy is bigger and there are only 17 parameters.. we can just loop over them.

azkrishpy (approved):

PR #63, opened 2026-05-25
Allow special characters in profile names.

Nevon:
*Description of changes:*
The AWS CLI and SDKs allows a wide set of characters in profile names. Trying to use the aws-crt credential provider with a profile named for example "foo.bar@123456" fails with a warning log like:

```
Profile declaration contains invalid characters: ".bar@123456"
```

Looking around at the SDKs and CLI, I can see two options:

* Allow "anything". Botocore falls into this category, as far as I can tell.
* Allow `A-Za-z0-9_-/.%@:+`. This is the case for at least the Typescript and Java SDKs (the Java implementation is in an open PR), with the Java one referencing "the AWS SDK shared-configuration SEP", which I'm assuming is some AWS internal standard.

This changes the profile name parsing to align with the SDKs. The one difference is that aws-crt historically allowed `\`, which the SDKs don't. I'm keeping it to remain backwards compatible.

### DIFF
--- a/source/aws_profile.c
+++ b/source/aws_profile.c
@@ -59,7 +59,8 @@ static bool s_is_identifier(uint8_t value) {
 
     if ((value_as_char >= 'A' && value_as_char <= 'Z') || (value_as_char >= 'a' && value_as_char <= 'z') ||
         (value_as_char >= '0' && value_as_char <= '9') || value_as_char == '\\' || value_as_char == '_' ||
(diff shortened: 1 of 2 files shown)


TingDaoK:
thank you for detailed investigation and contribution! 
The code has been merged with https://github.com/awslabs/aws-c-sdkutils/pull/65
Closing the PR.

PR #64, opened 2026-06-02
feat:  Add streaming aws-chunked content encoding decoder

TingDaoK:
*Issue #, if available:*

*Description of changes:*
  
  Adds a general-purpose streaming decoder for the aws-chunked content encoding format.
  
  The decoder is a push-style state machine that processes arbitrary-sized input buffers
  incrementally, outputting decoded data bytes without buffering the full response.
  
  API:
  - aws_chunked_decoder_new(options) — create decoder with optional trailer callback and expected content length
  - aws_chunked_decoder_process(decoder, input, output_buf) — feed bytes in, get decoded data out
  - aws_chunked_decoder_is_done(decoder) — check if terminal chunk + trailers fully parsed
  - aws_chunked_decoder_destroy(decoder) — clean up
  
  Key properties:
  - Zero allocations on the hot path (chunk data passthrough)
  - Scratch buffer lazily allocated only when metadata lines split across calls
  - Fail-hard on malformed input (permanent error state)
  - Trailer delivered via callback (supports multiple trailers)
  - Optional content length verification at terminal chunk
  - Single-threaded, not thread-safe
  
  Testing:
  - Data-driven test vectors in JSON (shareable across SDK implementations)
  - All success vectors run through multiple split patterns (all-at-once, 1 byte, 2 bytes)
  - Error vectors for invalid hex, missing CRLF, malformed trailer, length mismatch, data after done
  - Fuzz tests for crash/leak detection under random and multi-split input


include/aws/sdkutils/aws_chunked_decoder.h
```diff
@@ -0,0 +1,67 @@
...
+ */
+typedef int(
+    aws_chunked_decoder_on_trailer_fn)(struct aws_byte_cursor name, struct aws_byte_cursor value, void *user_data);
+
+struct aws_chunked_decoder_options {
+    struct aws_allocator *allocator;
+    aws_chunked_decoder_on_trailer_fn *on_trailer;
+    void *user_data;
+    /* If non-NULL, decoder verifies total decoded bytes match this value. */
+    const uint64_t *expected_content_length;
```
DmitriyMusatkin: why is this a ptr?

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,407 @@
...
+    aws_chunked_decoder_on_trailer_fn *on_trailer;
+    void *user_data;
+
+    /* Flags */
+    bool is_done;
+    bool has_error;
+};
+
+/* Forward declarations */
+static int s_state_chunk_size_line(
```
DmitriyMusatkin: state machine purely driven by pointers? dont know why but it seems scary and hard to debug? can we at least add some state enum?

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,407 @@
...
+
+/**
+ * Getline helper: scans input for a complete CRLF-terminated line, handling splits.
+ * Returns:
+ *  - AWS_OP_SUCCESS with *out_line set and *line_complete=true if a full line is available
+ *  - AWS_OP_SUCCESS with *line_complete=false if more data needed (partial buffered in scratch)
+ *  - AWS_OP_ERR on error (line too long)
+ *
+ * When line_complete=true, out_line points to the line content (without CRLF).
+ * If scratch was used, out_line points into scratch (caller must reset scratch after processing).
```
DmitriyMusatkin, line 81: how would user know if the scratch was used or not?

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,407 @@
...
+    struct aws_byte_buf *output_buf);
+
+struct aws_chunked_decoder {
+    struct aws_allocator *alloc;
+
+    /* State machine */
+    aws_chunked_decoder_state_fn *state;
+
+    /* Scratch buffer for partial metadata lines */
+    struct aws_byte_buf scratch;
```
DmitriyMusatkin, line 41: would it be simpler to just  have a static 1024 buffer here, rather than heap one?

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,407 @@
...
+            *out_line = aws_byte_cursor_from_buf(&decoder->scratch);
+            *line_complete = true;
+            return AWS_OP_SUCCESS;
+        }
+    }
+
+    /* Scan input for CRLF */
+    struct aws_byte_cursor found;
+    if (aws_byte_cursor_find_exact(input, &s_crlf, &found) == AWS_OP_SUCCESS) {
+        size_t crlf_pos = (size_t)(found.ptr - input->ptr) + 2;
```
DmitriyMusatkin: oof this logic with adding/subtracting seems kinda confusing. if you add 2 here it points just after crlf?

include/aws/sdkutils/aws_chunked_decoder.h
TingDaoK: because 0 is a valid expected length... 

And I want this option to be able to take 0 as a default to not check the decoded bytes.

source/aws_chunked_decoder.c
TingDaoK: user don't need to know.

source/aws_chunked_decoder.c
TingDaoK: arguable, mostly, I consider the scratch to be a concern case that only will be trigger if the http client sends part of one metadata lines (trailer or the size and signature), and those metadata lines are fairly small. 

I am a bit on the fence to allocate the scratch on init for a corner case.

But, if you feel strongly, I don't disagree.

include/aws/sdkutils/aws_chunked_decoder.h
TingDaoK: changed my mind, the expected content length should always be required, right? I'll update it to be uint64_t and requires it to be set. 0 is valid, but will fail the decode if any body decoded.

sbSteveK:
A more general error reporting related suggestion; we are using `AWS_ERROR_SDKUTILS_PARSE_FATAL` for wide spread of error types. The log that goes alongside the error describes exactly what is causing the error but I feel like we might want to have a subset of errors specific to chunked related parsing failures. Even if we don't add granular errors per error type, something along the lines of `AWS_ERROR_SDKUTILS_AWS_CHUNKED_PARSE_FATAL` could be useful to immediately isolate where an error is coming from.

Maybe this is a non-issue. I'm not sure how easily error logs are obtained when people using S3 run into them. If they're easily accessible (esp since they are error logs and not debug or trace) then this matters a lot less. If customers tend to only have the error code on hard to reproduce errors, then it becomes something we should really consider.

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,415 @@
...
+        decoder->is_done = true;
+        decoder->state = AWS_CHUNKED_DECODER_STATE_DONE;
+        return AWS_OP_SUCCESS;
+    }
+
+    /* Split on first ':' — value is everything after the first colon (may contain more colons).
+     * Per RFC 7230 § 3.2: header-field = field-name ":" OWS field-value OWS
+     * Trim optional whitespace from value only; field-name must not contain whitespace. */
+    struct aws_byte_cursor name = {0};
+    if (!aws_byte_cursor_next_split(&line, ':', &name) || name.len == line.len) {
```
sbSteveK, line 288: according to same RFC ( https://datatracker.ietf.org/doc/html/rfc7230#section-3.2 ) the `field-name = token` and a token must consist of one or more valid characters. Searching around this sounds like it's accurate and the most official documentation I can find related to a `token` is here: https://datatracker.ietf.org/doc/html/rfc2389#section-2.1 

We should just add a `name.len  == 0` as an additional or fail state in this if to cover for it.

sbSteveK: If we want to be super strict, we may also want to check that the name isn't all whitespace? That may be setup elsewhere though if we are trimming whitespace. Will have to dig further.

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,415 @@
...
+/*
+ * Parse a complete trailer line. Format: "<name>:<value>" or empty line signals end.
+ * Example: "trailer-key: abc123=="
+ * Empty line ("\r\n" with no content) means all trailers are done → transition to DONE.
+ */
+static int s_parse_trailer_line(struct aws_chunked_decoder *decoder, struct aws_byte_cursor line) {
+    if (line.len == 0) {
+        AWS_LOGF_TRACE(
+            AWS_LS_SDKUTILS_GENERAL,
+            "id=%p: decode complete, total bytes=%" PRIu64,
```
sbSteveK: Trivial: Maybe this is the ONLY decoder in sdkutils but even still, maybe we want to specify "aws chunked decode complete" here?

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,415 @@
...
+    struct aws_byte_cursor *input,
+    struct aws_byte_buf *output_buf) {
+
+    (void)output_buf;
+
+    while (input->len > 0 && decoder->chunk_processed < 2) {
+        uint8_t expected = (decoder->chunk_processed == 0) ? s_cr : s_lf;
+        if (input->ptr[0] != expected) {
+            AWS_LOGF_ERROR(
+                AWS_LS_SDKUTILS_GENERAL,
```
sbSteveK: The more I look at our logging in this, the more I feel like we should create a specific log subject for AWS_CHUNKED. Is there any reason not to add `AWS_LS_SDKUTILS_AWS_CHUNKED` to the `aws_sdkutils_log_subject` to narrow down both logs and provide more specific when they're generated within decode of aws chunked messages?

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,415 @@
...
+            decoder->scratch.len -= 1; /* remove trailing \r */
+            *out_line = aws_byte_cursor_from_buf(&decoder->scratch);
+            *line_complete = true;
+            return AWS_OP_SUCCESS;
+        }
+    }
+
+    /* Scan input for CRLF */
+    struct aws_byte_cursor found;
+    if (aws_byte_cursor_find_exact(input, &s_crlf, &found) == AWS_OP_SUCCESS) {
```
sbSteveK, line 104: Here we use `aws_byte_cursor_find_exact()` ( https://github.com/awslabs/aws-c-common/blob/48dd6cdff7bac0005a9c6b49c15a0765622c0e70/source/byte_buf.c#L306-L307 ) which has a substantial amount of unnecessary work (when used repeatedly at scale). If we know we are only checking for a specific two-byte pattern "\r\n" then it'd probably provide a significant improvement to do a manual search with a local helper function or even just in its place here with 
```
    const uint8_t *cr_pos = NULL;
    {
        const uint8_t *scan = input->ptr;
        const uint8_t *end = input->ptr + input->len;
        while (scan < end) {
            const uint8_t *cr = memchr(scan, '\r', end - scan);
            if (cr == NULL || cr + 1 >= end) {
                break;
            }
            if (cr[1] == '\n') {
                cr_pos = cr;
                break;
            }
            scan = cr + 1;
        }
    }

   // This check takes the place of the `aws_byte_cursor_find_exact() == AWS_OP_SUCCESS`
    if (cr_pos) {
        // we use the cr_pos and input->ptr instead of the found
        size_t line_len = cr_pos - input->ptr;
        // everything past here would be identical to the aws_byte_cursor_find_exact == AWS_OP_SUCCESS

```

This saves us struct copies, a number of guards, and aws_raise_error() calls when the exact search fails.

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,415 @@
...
+ */
+static int s_getline(
+    struct aws_chunked_decoder *decoder,
+    struct aws_byte_cursor *input,
+    struct aws_byte_cursor *out_line,
+    bool *line_complete) {
+    *line_complete = false;
+
+    /* Handle split CRLF: scratch ends with \r, input starts with \n */
+    if (decoder->scratch.len > 0 && decoder->scratch.buffer[decoder->scratch.len - 1] == s_cr) {
```
sbSteveK, line 92: Looking at the documentation here: https://chorus.aws.dev/doc/5o9wFqUXYjBR/aws-chunked-Terminal-Chunk-Format-Discrepancy it appears as though there's an extra blank line before the trailer. The example given in the doc is
```
0;chunk-signature=UNSIGNED-PAYLOAD\r\n
\r\n                                    <- extra blank line!
x-amz-wire-checksum-crc32:<base64>\r\n
\r\n
```
After the payload of 0 is processed, we transition to `AWS_CHUNKED_DECODER_STATE_TRAILER_LINE` where it will then process the `\r\n` "extra blank line!" which will set line_complete to true and out_line->len to 0. We then fall into `s_parse_trailer_line()` where `decoder->is_done` is set to true. There is more inc after this which will set the decoder to an error state.

I could be missreading/missinterpreting the doc and maybe we don't have to deal with the extra blank line but I'm calling it out in case we need to add a test that has this blank line to see if we handle it properly. 

If the spec is that there is ALWAYS an extra line before the trailers, we can probably just add a flag that flips to true after encountering the first blank line once we get to the AWS_CHUNKED_DECODER_STATE_TRAILER_LINE state.

source/aws_chunked_decoder.c
sbSteveK: This could be resolved by combining it with the comment related to creating a new log subject for chunky decoding.

source/aws_chunked_decoder.c
sbSteveK: This can be ignored. Bug on AWS Chunked end.

source/aws_chunked_decoder.c
TingDaoK: > which has a substantial amount of unnecessary work

I disagree. What substantial amount of unnescessary work you refer to here?

TingDaoK: in general, we should use a helper as much as possible to avoid accessing those pointer directly.

Accessing pointer directly is the recipe for a security nonsense. 

But, we should reset the error code when we are not finding `crlf` from the line

source/aws_chunked_decoder.c
TingDaoK: I am hesitate about this

I want to justify split on first `:`, reject if no `:` found with the RFC. That's the only structural requirement — without a colon, we can't produce a name/value pair at all.

`:value` splits fine into name="" + value="value". The aws-chunked trailer section isn't strictly HTTP headers — it's S3-specific framing that resembles them. We don't enforce tchar rules on names or field-content rules on values either. The caller (aws-c-s3) matches against specific expected names, so anything unexpected is simply ignored.

I'll leave this as-is — the decoder parses correctly, validation belongs to the consumer. unless you feel strongly otherwise

sbSteveK (approved):

source/aws_chunked_decoder.c
```diff
@@ -0,0 +1,424 @@
...
+    int rc = s_parse_chunk_size_line(decoder, line);
+    /* Done with this line, reset scratch which may have been holding it */
+    decoder->scratch.len = 0;
+    return rc;
+}
+
+/*
+ * Small helper to either do a static or dynamic append.
+ */
+static int s_output_append(struct aws_byte_buf *buf, const struct aws_byte_cursor *data) {
```
azkrishpy, line 203: there is a helper for this now? https://github.com/awslabs/aws-c-common/pull/1253

azkrishpy (approved):

azkrishpy (approved):

PR #68, opened 2026-06-26
addressing comments to bdd engine

azkrishpy:
*Issue #, if available:*

*Description of changes:*
Addressing comments from #62 

### DIFF
--- a/include/aws/sdkutils/private/endpoints_types_impl.h
+++ b/include/aws/sdkutils/private/endpoints_types_impl.h
@@ -129,6 +129,11 @@ struct aws_endpoints_parameter {
     bool is_deprecated;
     struct aws_byte_cursor deprecated_message;
     struct aws_byte_cursor deprecated_since;
+    /* 1-based index of its location in scope-value array.
+     * Although param_idx is 1-based, array has 0-based indexing
+     * index = 0 indicates the value has not yet been added to
+     * the register_map.
+     */
     size_t param_idx;
 };
 
@@ -227,6 +232,11 @@ struct aws_endpoints_rule_data_tree {
 struct aws_endpoints_condition {
     uint16_t expr_ref;
     struct aws_byte_cursor assign;
+    /* 1-based index of its location in scope-value array.
+     * Although assign_idx is 1-based, array has 0-based indexing
+     * index = 0 indicates the value has not yet been added to
+     * the register_map.
+     */
     size_t assign_idx;
 };
 
@@ -460,7 +470,7 @@ struct aws_endpoints_bdd_result {
 /* Max distinct named variables (parameters + condition assigns) per ruleset.
  * Scope values are stored in a fixed array of this size to avoid heap allocation
  * on the resolve hot path. Increase this constant if a larger ruleset is needed.
- * Current rulesets use ~20-30 slots. */
+ * 2026 July rulesets use ~40 slots. */
 enum { AWS_BDD_MAX_REGS = 128 };
 
 struct aws_bdd_scope {
--- a/include/aws/sdkutils/private/endpoints_types_impl.h
+++ b/include/aws/sdkutils/private/endpoints_types_impl.h
@@ -456,8 +456,13 @@ struct aws_endpoints_bdd_result {
     } data;
 };
 
+/* Max distinct named variables (parameters + condition assigns) per ruleset.
+ * Scope values are stored in a fixed array of this size to avoid heap allocation
+ * on the resolve hot path. The loader errors with AWS_ERROR_INVALID_ARGUMENT if
+ * exceeded. Increase this constant if a larger ruleset is needed.
+ * Current rulesets use ~20-30 slots. */
 enum {
-    s_max_regs = 128, /* max regs during eval */
+    s_max_regs = 128,
 };
 
 struct aws_endpoints_bdd_engine {
--- a/include/aws/sdkutils/private/endpoints_types_impl.h
+++ b/include/aws/sdkutils/private/endpoints_types_impl.h
@@ -459,16 +459,13 @@ struct aws_endpoints_bdd_result {
 
 /* Max distinct named variables (parameters + condition assigns) per ruleset.
  * Scope values are stored in a fixed array of this size to avoid heap allocation
- * on the resolve hot path. The loader errors with AWS_ERROR_INVALID_ARGUMENT if
- * exceeded. Increase this constant if a larger ruleset is needed.
+ * on the resolve hot path. Increase this constant if a larger ruleset is needed.
  * Current rulesets use ~20-30 slots. */
-enum {
-    s_max_regs = 128,
-};
+enum { AWS_BDD_MAX_REGS = 128 };
 
 struct aws_bdd_scope {
     struct aws_endpoints_bdd_engine *engine;
-    struct aws_endpoints_scope_value values[s_max_regs];
+    struct aws_endpoints_scope_value values[AWS_BDD_MAX_REGS];
 };
 
 struct aws_endpoints_bdd_engine_state {
--- a/include/aws/sdkutils/private/endpoints_types_impl.h
+++ b/include/aws/sdkutils/private/endpoints_types_impl.h
@@ -483,6 +483,16 @@ struct aws_endpoints_bdd_engine {
     struct aws_array_list conditions;
     struct aws_endpoints_condition *conditions_array_ptr;
 
+    /* This register map is a hash table to store indices of scope values array. In the previous engine,
+     * each scope had its own hash table of scope values which meant, every time we needed to
+     * update a value for a scope variable, we did a hash_table_find. Expensive. The new
+     * state stores it as a stack allocated scope array which would mean faster finding and
+     * replacing values. This works by storing register_map which would store a particular index
+     * at which that particular scope variable resides in the scope_values array and each assign
+     * scope variable and parameter get its index. if the index is 0 we can look up the correct
+     * index from the register map if not use the index we stored. Ideally we should have all the
+     * indices stored and ready at load time.
+     */
     struct aws_hash_table register_map;
 
     /* array of all exprs in the program. everything else indexes into this. */
--- a/include/aws/sdkutils/private/endpoints_types_impl.h
+++ b/include/aws/sdkutils/private/endpoints_types_impl.h
@@ -129,6 +129,7 @@ struct aws_endpoints_parameter {
     bool is_deprecated;
     struct aws_byte_cursor deprecated_message;
     struct aws_byte_cursor deprecated_since;
+    size_t param_idx;
 };
 
 struct aws_endpoints_ruleset {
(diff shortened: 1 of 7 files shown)


source/endpoints_bdd_engine.c
```diff
@@ -97,23 +102,25 @@ static int s_init_state(
     }
 
     /* Add defaults to the top level scope. */
-    for (struct aws_hash_iter iter = aws_hash_iter_begin(&engine->parameters); !aws_hash_iter_done(&iter);
-         aws_hash_iter_next(&iter)) {
-        const struct aws_byte_cursor key = *(const struct aws_byte_cursor *)iter.element.key;
-        struct aws_endpoints_parameter *value = (struct aws_endpoints_parameter *)iter.element.value;
+    for (size_t i = 0; i < aws_array_list_length(&engine->parameters); ++i) {
```
TingDaoK, line 100: why not move this load default to the loader?

source/endpoints_bdd_engine.c
```diff
@@ -97,23 +102,25 @@ static int s_init_state(
     }
```
TingDaoK, line 97: why we initailize the context before we apply the default from params? Should it be the context can override the default?

source/endpoints_bdd_engine.c
azkrishpy: two reasons:
- the number of parameters we have is fairly low (~17).
- there is no spec-mention of all required values will be provided with defaults. so we would have to copy context and then perform a check on required parameters anyway.

source/endpoints_bdd_loader.c
```diff
@@ -154,49 +151,58 @@ static int s_parse_one_parameter(
...
-            out_parameters,
-            allocator,
-            count,
-            aws_hash_byte_cursor_ptr,
-            aws_endpoints_byte_cursor_eq,
-            NULL,
-            s_callback_endpoints_parameter_destroy)) {
+    struct aws_endpoints_parameter *parameters =
+        aws_mem_calloc(allocator, count, sizeof(struct aws_endpoints_parameter));
+    if (!parameters) {
```
TingDaoK: don't check for allocation failure

source/endpoints_bdd_loader.c
```diff
@@ -154,49 +151,58 @@ static int s_parse_one_parameter(
...
+                AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE,
+                "Too many unique variables in ruleset. Increase s_max_regs (currently %d).",
+                s_max_regs);
+            return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        }
+        size_t param_idx = aws_hash_table_get_entry_count(&engine->register_map) + 1;
+        aws_hash_table_put(&engine->register_map, &param->name, (void *)param_idx, NULL);
+        param->param_idx = param_idx;
+    }
+
```
TingDaoK, line 163: instead of adding it to the table here, why not adding it in the for loop of `s_load_parameters`?

after parsing, just do hash table put with the name and the idx. since the `register_map` is empty at that time, you can just have the idx equal to the index of the param in the array. 

also, do we need to check if there is a duplicate params?

source/endpoints_bdd_loader.c
```diff
@@ -422,9 +434,16 @@ static int s_parse_one_condition(
         struct aws_hash_element *element = NULL;
         aws_hash_table_find(&engine->register_map, &cond->assign, &element);
         if (element != NULL) {
-            size_t reg_index = (size_t)element->value + 1;
+            size_t reg_index = (size_t)element->value;
             cond->assign_idx = reg_index;
         } else {
+            if (aws_hash_table_get_entry_count(&engine->register_map) >= s_max_regs) {
```
TingDaoK: it maybe simpler if we move this check after `cond->assign_idx = aws_hash_table_get_entry_count(&engine->register_map) + 1;`. Check for `cond->assign_idx` is valid or not

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -471,11 +477,22 @@ struct aws_endpoints_bdd_engine {
...
+     * each scope had its own hash table of scope values which meant, every time we needed to
+     * update a value for a scope variable, we did a hash_table_find. Expensive. The new
+     * state stores it as a stack allocated scope array which would mean faster finding and
+     * replacing values. This works by storing register_map which would store a particular index
+     * at which that particular scope variable resides in the scope_values array and each assign
+     * scope variable and parameter get its index. if the index is 0 we can look up the correct
+     * index from the register map if not use the index we stored. Ideally we should have all the
+     * indices stored and ready at load time.
+     */
     struct aws_hash_table register_map;
```
TingDaoK, line 509: in the doc, be clear what is the key and what is the value stored in the map.

source/endpoints_bdd_engine.c
```diff
@@ -31,15 +31,18 @@ static int s_copy_context_to_state(
...
-            idx = aws_hash_table_get_entry_count(&state->engine->register_map) + 1;
+        if (element == NULL) {
+            AWS_LOGF_ERROR(
+                AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE,
+                "Received a context variable not present in parameters: " PRInSTR,
+                AWS_BYTE_CURSOR_PRI(context_value->name.cur));
+            return aws_raise_error(AWS_ERROR_SDKUTILS_ENDPOINTS_RESOLVE_INIT_FAILED);
         }
 
+        size_t idx = (size_t)element->value - 1;
```
TingDaoK: i'd still suggest everytime we get the index from some map result, we should check the index is valid before we access the array with the index.

source/endpoints_bdd_engine.c
TingDaoK: to me everytime we resolve the endpoint with various context, currently we do:
1. copy every context to the state values
2. iterate all params ad copy every params has default value to the state (And if it's not set from the context)

Instead, I think to have the default state living with the engine will be more easier to follow, so that every resolve will be:
1. copy the default state values
2. apply every context to override/set the state values

source/endpoints_bdd_loader.c
azkrishpy: I was just replicating what we did for conditions. since we are loading parameters first not really, all parameters are unique.

source/endpoints_bdd_engine.c
azkrishpy: the idea was copy context first and if no context value provided add the default. no real difference in terms of performance I believe

source/endpoints_bdd_loader.c
azkrishpy: you're right count get_entry_count() only once would be ideal. switch this everywhere.

source/endpoints_bdd_loader.c
azkrishpy: ideally we can also do the same thing for conditions since we know the parameters count that was already loaded. worth considering?

source/endpoints_bdd_engine.c
azkrishpy: discussed, marking as resolved.

source/endpoints_bdd_engine.c
azkrishpy: we error out everywhere if we try to add an index beyond the defined bounds.. so unless the element itself is NULL, we should be getting the right index. if not we would be adding the bounds check in other places as well. Eg. Line 70.

If you still feel strongly about this I can add a bounds check.

TingDaoK: Yeah, I feel strongly about this

TingDaoK: The check is to protect accessing the array with possible security nonsense. In case we changing the map or anything, before we access the array, check against the boundary. Or make a helper to access the array. Or use our array.

source/endpoints_bdd_engine.c
```diff
@@ -97,39 +97,45 @@ static int s_init_state(
...
+        /* value should always be present in the register map since we load parameters into the map */
+        if (value->param_idx == 0) {
+            AWS_LOGF_ERROR(
+                AWS_LS_SDKUTILS_ENDPOINTS_RESOLVE,
+                "Value not present in register map: " PRInSTR,
+                AWS_BYTE_CURSOR_PRI(value->name));
+            return AWS_OP_ERR;
+        }
 
         /* Skip non-required values, since they cannot have default values. */
```
TingDaoK: trivial: add comments
```suggestion
        /* Ignore non-required values from the check. They cannot have default values. */
```

source/endpoints_bdd_loader.c
```diff
@@ -154,49 +153,52 @@ static int s_parse_one_parameter(
...
 
+    /* We are confident about not checking the register_map to see if the value
+     * already exists since parameters are the first variables to be loaded and
+     * guaranteed to be unique amongst themselves.
+     */
+    size_t param_idx = offset + 1;
+    AWS_ERROR_PRECONDITION3(
+        param_idx <= AWS_BDD_MAX_REGS,
+        AWS_ERROR_SDKUTILS_ENDPOINTS_UNSUPPORTED_RULESET,
+        "Too many unique variables in ruleset. Increase AWS_BDD_MAX_REGS.");
```
TingDaoK: trivial: for the loader, we can do one check at the end.

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -129,6 +129,7 @@ struct aws_endpoints_parameter {
     bool is_deprecated;
     struct aws_byte_cursor deprecated_message;
     struct aws_byte_cursor deprecated_since;
+    size_t param_idx;
```
TingDaoK, line 137: i feel it's worth to mention all those idx we stored are 1-based, but when we retrieve them from the value array, the array is 0 based. 

As I type this out, can we make the value array also 1-based? Or, it will just be confusing. I am fine either way, just want to bring it up

include/aws/sdkutils/private/endpoints_types_impl.h
```diff
@@ -456,8 +457,21 @@ struct aws_endpoints_bdd_result {
     } data;
 };
 
-enum {
-    s_max_regs = 128, /* max regs during eval */
+/* Max distinct named variables (parameters + condition assigns) per ruleset.
+ * Scope values are stored in a fixed array of this size to avoid heap allocation
+ * on the resolve hot path. Increase this constant if a larger ruleset is needed.
+ * Current rulesets use ~20-30 slots. */
```
TingDaoK:
```suggestion
 * 2026 July rulesets use ~20-30 slots. */
```

TingDaoK (approved):
all trivial

include/aws/sdkutils/private/endpoints_types_impl.h
azkrishpy: I think 0 is reserved for when we have not assigned an index yet and we want to look up the register_map for if the value has already been set. so i guess we can keep the convention. I have documented that both `assign_idx` and `param_idx` are 1-based.

